PR #21, opened 2020-09-17
riscv64: added opensbi bootloader

Maxez:
This PR adds opensbi bootloader for RISCV64 and introduces following changes:
1. moved riscv-pk submodule to riscv/riscv-pk (opensbi submodule is placed at the same level in riscv/opensbi),
2. added RISCV64 loader selection with RISCV_LOADER variable:
- RISCV_LOADER="bbl" -> use bbl as bootloader, generate .bbl kernel image,
- RISCV_LOADER="opensbi" -> use opensbi as bootloader, generate .osbi kernel image,
- RISCV_LOADER equals anything else -> no bootloader, use flat kernel binary .bin image,
Appropriate kernel image is saved in _boot/ folder when running build.sh script with 'image' or 'all' options ('image' option creates _boot/ folder).

### DIFF
--- a/_targets/build.project.riscv64-spike
+++ b/_targets/build.project.riscv64-spike
@@ -11,6 +11,8 @@ CROSS=riscv64-phoenix-
 #CFLAGS="-Os -Wall -g -fomit-frame-pointer -fdata-sections -ffunction-sections -mcmodel=medany -fno-builtin -Wstrict-prototypes"
 #LDFLAGS="-z max-page-size=0x1000" -fno-builtin
 
+export RISCV_LOADER="bbl"
+#export RISCV_LOADER="opensbi"
 export BUSYBOX_CONFIG=$(realpath "busybox_config")
 export DROPBEAR_CUSTOM_CFLAGS="-DLTC_NO_BSWAP"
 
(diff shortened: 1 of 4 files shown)


pawelpisarczyk:
Łukasz, target files for virt and spike should be separated because spike emulator doesn't implement uart16550

_targets/build.project.riscv64-spike
```diff
@@ -54,37 +56,51 @@ b_build_target() {
...
+		popd
+	fi
 }
 
 
 b_image_target() {
-	b_log "Creating image from $PREFIX_ROOTFS"
+	b_log "Creating image from ${PREFIX_ROOTFS}"
 	size=`find ${PREFIX_ROOTFS} -type f -exec du -k {} \; | awk 'BEGIN { sum=0 }; { sum+=$1; }; END { print sum }'`
 	size=`expr ${size} \* 150 / 100`
```
kemonats, line 90: better will be:
```
	size=$(find ${PREFIX_ROOTFS} -type f -exec du -k {} \; | awk 'BEGIN { sum=0 }; { sum+=$1; }; END { print sum }')
	size=$((size*150/100))
````

Comment (Maxez) [reply _targets/build.project.riscv64-spike:90]: I agree with using $() instead of `` ([here is why](http://mywiki.wooledge.org/BashFAQ/082)). I believe this problem deserves separate PR as it's present in many more places all around our codebase. I'll leave it for future fix.

**Comment by Maxez** [comment]
> Łukasz, target files for virt and spike should be separated because spike emulator doesn't implement uart16550

I restored separate file for virt target as requested.

File: _targets/build.project.riscv64-spike [hunk 2]
```diff
@@ -54,37 +56,51 @@ b_build_target() {
...
-
-	genext2fs -b $size -d $PREFIX_ROOTFS $PREFIX_BUILD/disk.ext2
-	cp $PREFIX_BUILD/disk.ext2 $PREFIX_BOOT/phoenix-riscv64-spike.disk
-
+	echo "Filesystem size: ${size}"
+
+	genext2fs -b ${size} -d ${PREFIX_ROOTFS} ${PREFIX_BUILD}/disk.ext2
+	cp ${PREFIX_BUILD}/disk.ext2 ${PREFIX_BOOT}/phoenix-${TARGET}.disk
+
+	if [ "${RISCV_LOADER}" == "riscv-pk" ]; then
```
nalajcie, line 96: use `=` instead of `==` for string comparison (`==` is bash extension, better to use POSIX sh compliant option) - this applies everywhere

_targets/build.project.riscv64-spike
```diff
@@ -54,37 +56,51 @@ b_build_target() {
...
+		pushd riscv/riscv-pk/build-${TARGET}
+			if [ "X${CLEAN}" == "Xclean" ]; then
+				#make distclean
+				../configure --host=${CROSS%?} --with-payload=${PREFIX_PROG}/phoenix-${TARGET}.elf --disable-fp-emulation
+			fi
+
+			env -i PATH="$PATH" make ${MAKEFLAGS} all
+		popd
+	elif [ "${RISCV_LOADER}" == "opensbi" ]; then
+		pushd riscv/opensbi/
```
nalajcie, line 76: build opensbi out-of-tree in `_build/$TARGET`- make proper use of `make O=<build_directory>`

_targets/build.project.riscv64-spike
```diff
@@ -54,37 +56,51 @@ b_build_target() {
...
+			if [ "X${CLEAN}" == "Xclean" ]; then
```
nalajcie, line 77: it it would be inside `_build/$TARGET` - it won't be needed anymore (`clean` removes the whole directory)

_targets/build.project.riscv64-spike
```diff
@@ -54,37 +56,51 @@ b_build_target() {
...
 	size=`expr ${size} \* 150 / 100`
-	echo "Filesystem size: $size"
-
-	genext2fs -b $size -d $PREFIX_ROOTFS $PREFIX_BUILD/disk.ext2
-	cp $PREFIX_BUILD/disk.ext2 $PREFIX_BOOT/phoenix-riscv64-spike.disk
-
+	echo "Filesystem size: ${size}"
+
+	genext2fs -b ${size} -d ${PREFIX_ROOTFS} ${PREFIX_BUILD}/disk.ext2
+	cp ${PREFIX_BUILD}/disk.ext2 ${PREFIX_BOOT}/phoenix-${TARGET}.disk
```
nalajcie, line 94: why not mv? it would be much quicker

nalajcie (changes requested):
Please rebase this PR onto the current master to update it with recent changes. Some additional comments added in the code.

_targets/build.project.riscv64-spike
Maxez: Fixed.

_targets/build.project.riscv64-spike
Maxez: Fixed.

_targets/build.project.riscv64-spike
Maxez: Fixed. I agree with using mv here since disk.ext2 partition is generated in the very same function (in the previous line). However, in general I prefer to copy the final artifacts and here is why:
1. Imagine that someone has built final image with "clean all" options passed in to build.sh script.
2. The person plays with resulting image and breaks the ext2 partition.
3. Rebuilding the partition by passing "image" option to build.sh should suffice (it wouldn't if the disk was generated in other function and we would use mv instead of cp).

_targets/build.project.riscv64-spike
Maxez: Removed "the clean part" - building out-of-tree as suggested.

_targets/build.project.riscv64-spike
nalajcie: I don't think I understand how `cp` instead `mv` would help in case user would break the ext2 partition. If we're considering user breaking something, he can break any file and using just `image` build target will result in unbootable image.

`mv` as opposed to `cp` is atomic (assuming operations within the same partition) so it actually reduces number of potential problems.

I think it's not all that important regarding this PR and can be addressed separately in the future.

nalajcie (approved):

nalajcie:
LGTM, but unfortunately You need to rebase once more because of merging https://github.com/phoenix-rtos/phoenix-rtos-project/pull/31

PR #26, opened 2020-10-20
Build enhancements

nalajcie:
**WORK IN PROGRES | DO NOT MERGE**

This PR includes many buildscript enhancements and fixes to make it possible building with docker environment.

This also adds github workflow to automatically test building for supported targets (experimental).

### DIFF
--- a/_targets/build.project.armv7a7-imx6ull
+++ b/_targets/build.project.armv7a7-imx6ull
@@ -5,8 +5,6 @@
 # Author: Kaja Swat, Aleksander Kaminski, Pawel Pisarczyk, Lukasz Kosinski
 #
 
-set -e
-
 CROSS=arm-phoenix-
 CFLAGS="-Os -Wall -Wstrict-prototypes -g -mcpu=cortex-a7 -mtune=cortex-a7 -mfloat-abi=hard -mthumb\
 	-fomit-frame-pointer -mno-unaligned-access -fdata-sections -ffunction-sections -fno-builtin"
@@ -14,7 +12,7 @@ CFLAGS="-Os -Wall -Wstrict-prototypes -g -mcpu=cortex-a7 -mtune=cortex-a7 -mfloa
 
 LDFLAGS="-z max-page-size=0x1000"
 
-export BUSYBOX_CONFIG=$(realpath "busybox_config")
+export BUSYBOX_CONFIG="$(realpath "busybox_config")"
 export DROPBEAR_CUSTOM_CFLAGS="-DLTC_NO_BSWAP"
 
 #
@@ -38,16 +36,19 @@ export PORTS_CURL=y
 b_build_target() {
 	mkdir -p "$PREFIX_ROOTFS"/dev "$PREFIX_ROOTFS"/local "$PREFIX_ROOTFS"/data "$PREFIX_ROOTFS"/mnt "$PREFIX_ROOTFS"/var
 
-	ln -f $PREFIX_ROOTFS/bin/{psh,mount}
-	ln -f $PREFIX_ROOTFS/bin/{psh,ps}
-	ln -f $PREFIX_ROOTFS/bin/{psh,mem}
-	ln -f $PREFIX_ROOTFS/bin/{psh,perf}
-	ln -f $PREFIX_ROOTFS/bin/{psh,sync}
-	ln -f $PREFIX_ROOTFS/bin/{psh,reboot}
+	# TODO: move to psh install target
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,mount}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,bind}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,ps}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,mem}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,perf}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,sync}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,reboot}
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,top}
 
 	b_log "Creating CPIO archive"
-	pushd $PREFIX_PROG_STRIPPED >/dev/null
-		ls psh dummyfs imx6ull-flash imx6ull-uart psd | cpio -H newc -ov >$PREFIX_BUILD/phoenix-rtos-kernel/programs.cpio 2>/dev/null
+	pushd "$PREFIX_PROG_STRIPPED" >/dev/null || exit 1
+		ls psh dummyfs imx6ull-flash imx6ull-uart psd | cpio -H newc -ov > "$PREFIX_BUILD/phoenix-rtos-kernel/programs.cpio" 2>/dev/null
 	popd >/dev/null
 
 	# Relink kernel to include newly created cpio archive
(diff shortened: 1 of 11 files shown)


nalajcie:
Actually, this will probably be changed tomorrow to branch instead of pull request, as github actions configured in pull request are not triggered.

Unfortunately this PR introduces CI for 'push' and 'pull_request' only on main repo, it will look like that:
https://github.com/nalajcie/phoenix-rtos-project/pull/1

_targets/build.project.armv7a7-imx6ull
```diff
@@ -57,20 +58,19 @@ b_build_target() {
...
+	# FIXME: does work only on macOS, for now hackish solution based on extending PATH
 	mtd-utils/build.sh
+	PATH="$(pwd)/_build/host/prog/:$PATH"
+	IMG="$PREFIX_BOOT/phoenix-armv7a7-imx6ull.jffs2"
 
-	_build/host/prog/mkfs.jffs2 -n -U -m none -e $((64*4096)) -s 4096 -r "$PREFIX_ROOTFS"/ -o _boot/phoenix-armv7a7-imx6ull.jffs2
+	mkfs.jffs2 -n -U -m none -e $((64*4096)) -s 4096 -r "$PREFIX_ROOTFS"/ -o "$IMG"
 
-	sz=`du -k _boot/phoenix-armv7a7-imx6ull.jffs2 | awk '{ print $1 }'`
+	sz=$(du -k "$IMG" | awk '{ print $1 }')
```
kemonats, line 74: maybe `cut` instead `awk`
```
sz=$(du -k "$IMG" | cut -f1)
```

build.project
```diff
@@ -7,22 +7,40 @@
...
 #TARGET=armv7a7-imx6ull
 #TARGET=ia32-generic
 #TARGET=riscv64-spike
-TARGET=riscv64-virt
+#TARGET=riscv64-virt
+
+# optional args will be printed in red
+b_list_available_targets() {
+    [ "$#" -gt 0 ] && echo -e "\e[1;31m$*\e[0m"
+    echo -e "Please specify ia valid traget by setting TARGET variable to one of:"
```
kemonats, line 23: typo
```
echo -e "Please specify a valid target by setting TARGET variable to one of:"
```

build.project
```diff
@@ -7,22 +7,40 @@
...
+
+# it's best to provide TARGET by env variable
+if [ -z "$TARGET" ]; then
+    b_list_available_targets "TARGET variable not set"
+    exit 1
+fi
+
 
+TARGET_FAMILY=$(echo "$TARGET" | awk -F- '{ print $1 }')
+TARGET_SUBFAMILY="$TARGET_FAMILY-$(echo "$TARGET" | awk -F- '{ print $2 }')"
```
kemonats, line 35: This can be simpliefied to:
```
TARGET_FAMILY=$(echo "$TARGET" | cut -d- -f1)
TARGET_SUBFAMILY=$(echo "$TARGET" | awk -F- '{ print $1"-"$2 }')
```
or without creating processes:
```
TARGET_FAMILY=${TARGET%%-*}
tmp=${TARGET#*-}
TARGET_SUBFAMILY="$TARGET_FAMILY-${tmp%%-*}"
unset tmp
```

build.project
nalajcie: Thanks, fixed.

build.project
nalajcie: Yeah, I wanted to do something with that ugly one-liner but ended with using `printf` in `awk` - which was not better by any means.

I think we can use string substitution as You suggested (as it's POSIX sh compliant), but I would prefer to do it separately from this PR (probably introduce it as a function to `phoenix-rtos-build/build.subr`?).

For now I've used Your version of `awk` script as it's tidier than the previous solution.

_targets/build.project.armv7a7-imx6ull
```diff
@@ -38,16 +36,19 @@ export PORTS_CURL=y
 b_build_target() {
 	mkdir -p "$PREFIX_ROOTFS"/dev "$PREFIX_ROOTFS"/local "$PREFIX_ROOTFS"/data "$PREFIX_ROOTFS"/mnt "$PREFIX_ROOTFS"/var
```
kemonats, line 39: shorter
```
	mkdir -p "$PREFIX_ROOTFS/"{dev,local,data,mnt,var}
```

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -38,49 +36,53 @@ export PORTS_CURL=y
 b_build_target() {
 	mkdir -p "$PREFIX_ROOTFS"/dev "$PREFIX_ROOTFS"/local "$PREFIX_ROOTFS"/data "$PREFIX_ROOTFS"/mnt "$PREFIX_ROOTFS"/var
```
kemonats, line 39: shorter
```
	mkdir -p "$PREFIX_ROOTFS/"{dev,local,data,mnt,var}
```

_targets/build.project.ia32-generic
```diff
@@ -39,35 +37,38 @@ b_build_target() {
 
 	mkdir -p "$PREFIX_ROOTFS"/dev "$PREFIX_ROOTFS"/local "$PREFIX_ROOTFS"/data "$PREFIX_ROOTFS"/mnt "$PREFIX_ROOTFS"/var
```
kemonats, line 40: shorter
```
 mkdir -p "$PREFIX_ROOTFS/"{dev,/local,data,mnt,var}
```

_targets/build.project.riscv64-spike
```diff
@@ -70,21 +70,20 @@ b_build_target() {
...
 
 	cp riscv-pk/build-${TARGET}/bbl $PREFIX_BOOT/phoenix-riscv64-spike.bbl
 }
 
 
 b_image_target() {
 	b_log "Creating image from $PREFIX_ROOTFS"
-	size=`find ${PREFIX_ROOTFS} -type f -exec du -k {} \; | awk 'BEGIN { sum=0 }; { sum+=$1; }; END { print sum }'`
-	size=`expr ${size} \* 150 / 100`
+	size=$(find "${PREFIX_ROOTFS}" -type f -exec du -k {} \; | awk 'BEGIN { sum=0 }; { sum+=$1; }; END { print sum }')
```
kemonats, line 84: Instead of call `du` on each file, maybe it will be better:
```
size=$(find "${PREFIX_ROOTFS}" -type f -print0 | du -c -k --files0-from - | awk 'END{print $1}')
```
You can do arithmetic in awk:
```
awk 'END{print($1 * 150) / 100}'
```

_targets/build.project.armv7a7-imx6ull
nalajcie: Not a topic for this PR, but fixed anyway - IMHO creating empty dirs should be done as a part of `root-skel` (this is part of the skeleton, but keeping empty dirs in git repo is problematic) - so fixed globally in `phoenix-rtos-build/build.sh` as You've suggested

Fixed in: https://github.com/phoenix-rtos/phoenix-rtos-build/commit/a7185003e33388cd075a2a6e32c78ad2d6c7bc23

_targets/build.project.armv7m7-imxrt106x
nalajcie: Not a topic for this PR, but fixed anyway - IMHO creating empty dirs should be done as a part of `root-skel` (this is part of the skeleton, but keeping empty dirs in git repo is problematic) - so fixed globally in `phoenix-rtos-build/build.sh` as You've suggested

Fixed in: https://github.com/phoenix-rtos/phoenix-rtos-build/commit/a7185003e33388cd075a2a6e32c78ad2d6c7bc23

_targets/build.project.riscv64-spike
nalajcie: I think leaving it as is is better - the first line computes the real sum of file sizes and the second just estimates the partition size to be 1.5x larger than that.

Maybe it would increase readability to change variables naming to `fileSizeSum` and `partSize`, but it can be done in a separate PR if desirable.

kemonats: What do you think about the use of find and du as proposed.

nalajcie: You're right - I didn't dig into the ext2 image creation too much, the main point of this PR is to make all targets compilable, introduce docker-based build and basic CI using Github Actions.

The ext2 creation needs to be probably revised, as `genext2fs` takes size `-b` in blocks, and we should take into account that block size on host filesystem (eg. Your EXT4) might be different than on the target image.

For now I think we don't have the ability to mount ext2 partition on riscv64 targets, it would be better to focus on ia32-generic on which mounting ext2 works (can be tested under qemu).

This is not related to the buildsystem enhancements, should be done as a separate PR.

nalajcie: edit: I've checked that default block size in `genext2fs` is 1024, so we should probably get the ceil of file_size / 1024, eg:
```bash
echo $((($(stat -c "%s" file.txt) + 1023) / 1024))
```

But I'm not proficient enough to make the change and currently don't have the means to test it, so I'm leaving it as-is.

I think @Maxez is more knowledgeable in this topic, maybe our comments would help him.

PR #33, opened 2020-11-05
imxrt106x: copy plo-ram*.img to _boot; clean up script.

HBuczynski:
Majority of the changes in this script were made after checking correctness using shellcheck app.

### DIFF
--- a/_targets/build.project.armv7m7-imxrt106x
+++ b/_targets/build.project.armv7m7-imxrt106x
@@ -39,7 +39,7 @@ export PORTS_CURL=y
 #
 SIZE_PAGE=$((0x200))
 PAGE_MASK=$((0xfffffe00))
-KERNEL_OFFS=$((0x9000))
+KERNEL_OFFS=$((0x11000))
 
 
 #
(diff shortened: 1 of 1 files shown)


agkaminski (approved):
Looks good to me, but I am no bash expert

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -95,34 +94,36 @@ b_build_target() {
...
 }
 
 
 b_add2img() {
-	printf "Copying %s (offs=%dB)\n" $1 $OFFSET
-	dd if=$1 of=$2 bs=1 seek=$OFFSET conv=notrunc 2>/dev/null
-	SZ=$((($(stat -c "%s" $1) + $SIZE_PAGE - 1)&$PAGE_MASK))
-	OFFSET=$(($OFFSET+$SZ))
+	printf "Copying %s (offs=%dB)\n" "$1" "$OFFSET"
+	dd if="$1" of="$2" bs=1 seek="$OFFSET" conv=notrunc 2>/dev/null
```
nalajcie, line 104: copying byte-by-byte is not optimal, You should at least use `SIZE_PAGE` block size

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -95,34 +94,36 @@ b_build_target() {
...
+	SZ=$((($(stat -c "%s" "$1") + SIZE_PAGE - 1) & PAGE_MASK))
```
nalajcie, line 105: You're doing the same thing at least 3 times (lines:57, 61, 105), it would be better to make it a function (like `file_size_rounded` or similar)

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -95,34 +94,36 @@ b_build_target() {
...
+	OFFSET=$((OFFSET + SZ))
 }
 
 
 b_image_target() {
 	b_log "Creating image"
 
 	IMG=_boot/phoenix-${TARGET}.disk
```
nalajcie, line 113: Is this a NOR FLASH image? If so, I think:
 - all unused bytes should be initialized to `0xFF`
 - the flash image should have fixed size (equal to NOR partition/whole flash size)

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -95,34 +94,36 @@ b_build_target() {
...
 	# Kernel
-	OFFSET=$KERNEL_OFFS
-	b_add2img ${PREFIX_PROG_STRIPPED}phoenix-${TARGET}.elf $IMG
+	OFFSET="$KERNEL_OFFS"
+	b_add2img "${PREFIX_PROG_STRIPPED}phoenix-${TARGET}.elf" "$IMG"
 
 	#Programs
 	for prog in "${PROGS[@]}"; do
-		b_add2img ${PREFIX_PROG_STRIPPED}$prog $IMG
+		b_add2img "${PREFIX_PROG_STRIPPED}$prog" "$IMG"
```
nalajcie, line 126: it seems prone to errors that You compute where the programs will land in flash two times (first for PLO, then for actual placement). It would be better to put images into the image and write the PLO script at the same time.

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -48,24 +48,23 @@ KERNEL_OFFS=$((0x9000))
...
-		((poffs+=$psz))
+		psz=$((($(stat -c "%s" "${PREFIX_PROG_STRIPPED}$prog") + SIZE_PAGE - 1) & PAGE_MASK))
+		printf "@%s(%x:%x)\n" "$prog" "$poffs" "$psz" >> "$PREFIX_BUILD/plo/script.plo"
+		((poffs+=psz))
 	done
 
-	cat _targets/armv7m7-imxrt106x.plo >> $PREFIX_BUILD/plo/script.plo
-	printf "\0" >> $PREFIX_BUILD/plo/script.plo
+	cat _targets/armv7m7-imxrt106x.plo >> "$PREFIX_BUILD/plo/script.plo"
+	printf "\0" >> "$PREFIX_BUILD/plo/script.plo"
```
nalajcie, line 67: just a quick note - appending the "images placement description" to the `plo` can't be the final solution as we won't be able to change them during FW upgrade (without overwriting PLO - and this is something you **never** want to do in the field). I'm assuming the current solution would evolve in that direction in the future.

nalajcie (commented):
Thanks for fixing many SH warnings/errors.

Sorry I wasn't able do the review before Paweł merged it - even so I've put some general remarks which may be addressed in the following commits.

PR #36, opened 2020-11-20
build.project: allow to run b_build_project function from specific target build file

HBuczynski:
This PR should be merged after approving

- https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/155

- https://github.com/phoenix-rtos/phoenix-rtos-build/pull/34

- https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/74

### DIFF
--- a/build.project
+++ b/build.project
@@ -3,25 +3,18 @@
 # Shell script for building Phoenix-RTOS microdistribution
 #
 # Copyright 2018, 2019, 2020 Phoenix Systems
-# Author: Kaja Swat, Aleksander Kaminski, Pawel Pisarczyk, Lukasz Kosinski
+# Author: Kaja Swat, Aleksander Kaminski, Pawel Pisarczyk, Lukasz Kosinski, Hubert Buczynski
 #
 [ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
 
-#TODO: remove
-#TARGET=armv7m4-stm32l4x6
-#TARGET=armv7m7-imxrt105x
-#TARGET=armv7m7-imxrt106x
-#TARGET=armv7m7-imxrt117x
-#TARGET=armv7a7-imx6ull
-#TARGET=ia32-generic
-#TARGET=riscv64-spike
-#TARGET=riscv64-virt
+
 
 # optional args will be printed in red
 b_list_available_targets() {
     [ "$#" -gt 0 ] && echo -e "\e[1;31m$*\e[0m"
     echo -e "Please specify a valid traget by setting TARGET variable to one of:"
-    ls -1 _targets/build.project.* | cut -d. -f 3
+    find _targets/build.project.*  | cut -d. -f 3
+    find _projects/build.project.* 2>/dev/null | cut -d. -f 3
 }
 
 # it's best to provide TARGET by env variable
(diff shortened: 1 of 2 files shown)


build.project
```diff
@@ -30,21 +23,39 @@ if [ -z "$TARGET" ]; then
...
 
-BUILD_TARGET_FILE="_targets/build.project.${TARGET_SUBFAMILY}"
 if [ ! -f "$BUILD_TARGET_FILE" ]; then
     b_list_available_targets "TARGET='$TARGET' is unavailable for this project"
     exit 1
 fi
 
 . "$BUILD_TARGET_FILE"
 
+
```
nalajcie, line 43: Non-intuitive logic flow, I would write:
```bash
if [ -n "$TARGET_PROJECT" ]; then
    if [ ! -f "$BUILD_PROJECT_FILE" ]; then
        b_list_available_targets "TARGET='$TARGET' is unavailable for this project"
        exit 1
    fi

    . "$BUILD_PROJECT_FILE"
fi
```

The main logic is to include the file **if we have project defined**, not if the file exists.

I'm not even sure why checking the last char of `$TARGET` variable is even needed in Your code.

build.project
```diff
@@ -30,21 +23,39 @@ if [ -z "$TARGET" ]; then
...
+if [ -f "$BUILD_PROJECT_FILE" ]; then
+	. "$BUILD_PROJECT_FILE"
+else
+	if [ ! -z "$TARGET_PROJECT" ] || [ "${TARGET: -1}" == "-" ]; then
```
nalajcie, line 47: it's always better to use "=" instead of "==" for string comparison in sh ("==" is bash extension, "=" is POSIX sh compliant)

nalajcie (changes requested):
Overall, the changes are good, I would simplify the logic for includeing the project-specific file.

build.project
HBuczynski: Ok, the logic flow can be changed.

Checking the last character covers the case when user defines `TARGET` as follow:
`armv7m7-imxrt106x-`
Then:
- `$TARGET_PROJECT`  is empty 
- `$TARGET` contains wrong char at the end
- in your proposition this mistake will not be caught

nalajcie: Ok, You're talking about setting for example `TARGET="ia32-generic-"`.

This is not a problem related to the existence of `$BUILD_PROJECT_FILE`, so it should not be checked there. 

Currently, having `-` at the end of `TARGET` won't break anything, but we can check it nevertheless. The other approach would be to always translate `TARGET` into triplet (eg. `ia32-generic` => `ia32-generic-none`) - that way it's handled in the autotools/automake (eg. when we compile `gcc`, the `arm-phoenix` target is translated to `arm-none-phoenix`).

I would stick with checking format of `TARGET` variable if You feel it's needed - but checking just the suffix might not be enough (eg. it won't catch `ia32-generic--project` as an error). Maybe we can check the whole format:

```bash
if [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY" ] || [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY-$TARGET_PROJECT" ]; then
[...]
```
but probably it would be better to rewrite the TARGET_FAMILY/SUBFAMILY/PROJECT  assignment to use switch-case with patterns and default case with error message.

HBuczynski: I've made changes as you proposed associated with checking the whole format. I think this verification is needed. Some projects has hardcode `TARGET` name as directory name.

nalajcie: ok, fine be me.

build.project
```diff
@@ -30,21 +23,44 @@ if [ -z "$TARGET" ]; then
...
+TARGET_SUBFAMILY="$(echo "$TARGET" | awk -F- '{ print $2 }')"
+TARGET_PROJECT="$(echo "$TARGET" | awk -F- '{ print $3 }')"
+
+
+BUILD_TARGET_FILE="_targets/build.project.${TARGET_FAMILY}-${TARGET_SUBFAMILY}"
+BUILD_PROJECT_FILE="_projects/build.project.${TARGET_FAMILY}-${TARGET_SUBFAMILY}-${TARGET_PROJECT}"
+
+
+if [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY" ] || [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY-$TARGET_PROJECT" ]; then    
+	b_list_available_targets "TARGET='$TARGET' is unavailable for this project"
```
nalajcie, line 37: nitpicking, not a merge blocker: the error message should state that TARGET variable has invalid format, not that it's unavailable (we haven't checked that yet)

build.project
```diff
@@ -30,21 +23,44 @@ if [ -z "$TARGET" ]; then
...
-TARGET_SUBFAMILY="$(echo "$TARGET" | awk -F- '{ print $1"-"$2 }')"
+TARGET_SUBFAMILY="$(echo "$TARGET" | awk -F- '{ print $2 }')"
+TARGET_PROJECT="$(echo "$TARGET" | awk -F- '{ print $3 }')"
+
+
+BUILD_TARGET_FILE="_targets/build.project.${TARGET_FAMILY}-${TARGET_SUBFAMILY}"
+BUILD_PROJECT_FILE="_projects/build.project.${TARGET_FAMILY}-${TARGET_SUBFAMILY}-${TARGET_PROJECT}"
+
+
+if [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY" ] || [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY-$TARGET_PROJECT" ]; then
```
nalajcie, line 36: shit, my bad, I've done an error, You should put `&&` in here instead of `||`, otherwise it's always true

HBuczynski: Event putting `&&` for `$TARGET = armv7m7-imxrt106x-` it does not work

nalajcie: yeah, You're right, we should check if `TARGET_PROJECT` is non-empty. I propose more hackish approach:
```bash
if [ "$TARGET" != "$TARGET_FAMILY-$TARGET_SUBFAMILY${TARGET_PROJECT:+-$TARGET_PROJECT}" ]; then
```
but I don't know if the readability decrease is worth doing it that way

nalajcie: Your approach also works.

nalajcie (approved):

PR #37, opened 2020-11-25
submodules: update

HBuczynski:


nalajcie:
We can't update `phoenix-rtos-corelibs` submodule because of phoenix-rtos/phoenix-rtos-corelibs#1. I hope updating everything apart from it will suffice.

nalajcie (approved):

PR #50, opened 2021-03-03
Submodule update: plo

Cinek28:
Including change fixing uart rx error


agkaminski:
Can you connect it with some JIRA task?

Cinek28:
Created and added jira task label

PR #57, opened 2021-03-20
ia32-generic.sh: simpler checking of arguments

kemonats:
I changed the argument checking for:
`-virtio-gpu|-vgpu`
`-virtio-net|-vnet`
`-rtl8139|-rtl`

I think that using `case...esac` in this case will be better.
In `-vgpu` I assumed a range of numbers `1-9` ( this can be changed of course ).

For me `echo && command` looks strange.

The same applies to `riscv64-virt.sh`.

To the author:
- Help for `-rtl8139|-rtl` is missing.
- Maybe it would be better to use subcommands instead of many scripts.

### DIFF
--- a/scripts/qemu/ia32-generic.sh
+++ b/scripts/qemu/ia32-generic.sh
@@ -54,10 +54,12 @@ while [ "$#" -gt 0 ]; do
 		# Add VirtIO GPU device
 		-virtio-gpu|-vgpu)
 			GPUS+=" -device virtio-gpu-pci"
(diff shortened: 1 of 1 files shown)


Maxez (approved):
I agree with changes proposed by @kemonats. The scripts required more fixes so I created another PR that gathers all of them in one place (including changes from this PR): [link](https://github.com/phoenix-rtos/phoenix-rtos-project/pull/58). I'm closing this PR in favor of the second one.

PR #53, opened 2021-03-09
Added scripts to simplify running Phoenix-RTOS on QEMU (JIRA: PD-20)

Maxez:
Added scripts allow to easily run Phoenix-RTOS image under QEMU with additional devices and options (e.g. with attached virtio devices). The general idea is to create a wrapper scripts for QEMU startup command. Each QEMU emulated target (ia32-generic and riscv64-virt for now) has a main startup script named 'target.sh' (e.g. ia32-generic.sh for ia32-generic target). The main script extends default qemu startup command with additional options which simplify device creation (e.g. -vblk, -vgpu or -vnet). One can still pass in original qemu options like -cpu, -smp or -kernel. There are also additional scripts named 'target-dev.sh'  which attach specific device to default system configuration (e.g. ia32-generic-vblk.sh runs default ia32-generic Phoenix-RTOS image with extra virtio-blk devices - attached disk images should be passed in as arguments). 

Examples:
./ia32-generic.sh - run default configuration (the system boots up to plo from default target image), 
./ia32-generic.sh -kernel ../../_build/ia32-generic/prog/phoenix-ia32 - run provided kernel directly without plo (example of original qemu options passthrough),
./ia32-generic-ata.sh ../../_boot/phoenix-ia32-generic.disk - demonstrates adding extra ATA drive to the system,

PR #58, opened 2021-03-22
scripts/qemu: fixed bugs reported by shellcheck

Maxez:
JIRA: PD-20
This PR fixes bugs mentioned in [previous PR](https://github.com/phoenix-rtos/phoenix-rtos-project/pull/53). The scripts were checked with shellcheck.

### DIFF
--- a/scripts/qemu/ia32-generic-ata.sh
+++ b/scripts/qemu/ia32-generic-ata.sh
@@ -8,8 +8,8 @@
 
 # Add ATA drives
 while [ "$#" -gt 0 ]; do
-	DRIVES+=" -ata $1"
+	DRIVES+=("-ata" "$1")
 	shift
 done
 
-exec bash "$(dirname "$BASH_SOURCE")/ia32-generic.sh" $DRIVES
+exec bash "$(dirname "${BASH_SOURCE[0]}")/ia32-generic.sh" "${DRIVES[@]}"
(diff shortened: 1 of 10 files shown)


scripts/qemu/ia32-generic-ata.sh
```diff
@@ -8,8 +8,8 @@
 
 # Add ATA drives
 while [ "$#" -gt 0 ]; do
-	DRIVES+=" -ata $1"
+	DRIVES+=("-ata" "$1")
```
kemonats, line 11: Maybe it would be a good idea to declare `DRIVERS` as an array.
```
# Add ATA drives
DRIVERS=()
while [ "$#" -gt 0 ]; do
```

scripts/qemu/ia32-generic-ata.sh
```diff
@@ -8,8 +8,8 @@
 
 # Add ATA drives
 while [ "$#" -gt 0 ]; do
```
kemonats, line 10: For varable `$#` double quote is not needed ( it is a number )

scripts/qemu/ia32-generic-ata.sh
Maxez: Added in next commit.

scripts/qemu/ia32-generic-ata.sh
Maxez: Added in next commit.

PR #63, opened 2021-04-19
tests: add 'host-pc' target

jsarzy:


.github/workflows/ci.yml
```diff
@@ -73,6 +73,10 @@ jobs:
           name: phoenix-rtos-${{ matrix.target }}
           path: _boot
 
+      - name: Fix file permissions
+        if: ${{ matrix.target == 'host-pc' }}
+        run: find _boot/host-pc/ -type f -exec chmod 755 {} \;
```
jsarzy, line 78: Action upload-artifact do not preserve file's permissions.

.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/phoenix-rtos/libphoenix
 [submodule "phoenix-rtos-build"]
 	path = phoenix-rtos-build
-	url = https://github.com/phoenix-rtos/phoenix-rtos-build
+	url = https://github.com/jsarzy/phoenix-rtos-build
```
jsarzy, line 9: Need to be changed before merge.

.gitmodules
```diff
@@ -27,7 +27,7 @@
 	url = https://github.com/phoenix-rtos/phoenix-rtos-posixsrv
 [submodule "phoenix-rtos-tests"]
 	path = phoenix-rtos-tests
-	url = https://github.com/phoenix-rtos/phoenix-rtos-tests
+	url = https://github.com/jsarzy/phoenix-rtos-tests
```
jsarzy, line 30: Need to be changed before merge.

_targets/build.project.host-pc
```diff
@@ -17,5 +16,13 @@ b_image_target() {
     :
 }
 
+b_test_target() {
+	b_log "Installing tests in the phoenix-rtos filesystem"
+	(cd phoenix-rtos-tests && make $MAKEFLAGS install)
+
+	b_log "Copying binaries into $PREFIX_BOOT/$TARGET"
```
nalajcie, line 23: I would go for maybe creating host-pc "rootfs image" as a `.tar` file (in the `image` stage) as some test may need additional input/output files in the future. This would also preserve the file permissions and case sensitive files in case it would be needed.

jsarzy: I don't know if I understand correctly the idea. Do we want to pack single test as a tar or "taring" the whole rootfs?
Now `test` stage is executed before `image`.

nalajcie: I've left the comment on the other PR. Basically I would see the process on host-pc similar to process on other platforms:
 - compile all necessary components, install basic components to rootfs (build targets: `core project`...)
 - install tests to rootfs (build targets: `test`, inside `b_test_target`)
 - create rootfs image (build targets: `image`) - for some targets it would be JFFS2/EXT partition image, for host-pc it could be `rootfs.tar`

During testing:
 - setup the device (reset, clean, flash with new rootfs/image) -> for qemu this is no-op, for host-pc this is unpacking/mounting rootfs, for hw targets it will be target specific
 - execute tests sequentially

As I've written in other PR, currently there is no real advantage with the `rootfs.tar` approach (it would still make us use extra step in the workflow) so the current solution is enough for now. We may address it again when we will make changes to the runner to be able to support HW targets.

PR #66, opened 2021-04-26
Fix imx6ull building and image creation

MaciejPurski:


pawelpisarczyk (commented):
Maciej, I'm not sure if this is a good idea to disable OpenVPN compilation. You can do this by defining variable in specific project, but I think that in generic project these ports should be included in the image.

MaciejPurski:
> Maciej, I'm not sure if this is a good idea to disable OpenVPN compilation. You can do this by defining variable in specific project, but I think that in generic project these ports should be included in the image.

I'm ok with that. So I'll get rid of the first commit.
For now, I don't understand, why the CI build fails on zynq7000.

build.project
```diff
@@ -69,6 +69,7 @@ b_build() {
 #
 b_image() {
 	b_image_target
+	declare -F b_image_project && b_image_project
```
nalajcie, line 72: this returns 1 which makes build.sh exit with return code 1

MaciejPurski: Ok. In the next version I'l try something like this:
```suggestion
	declare -F b_image_project && b_image_project
	return 0
```

PR #68, opened 2021-04-29
ports: MicroPython port for Phoenix-RTOS

niewim19:
added enviromental variable defining if micropython is compiled in ia32 target and added missing symlink for pshlogin applet

JIRA: BES-145 BES-104


nalajcie:
please make it 2 separate atomic commits (making unrelated changes in single commit is considered bad practice).

Please consider also other enhancements to micropython in `phoenix-rtos-ports`:
- unpack the patches in `phoenix-rtos-ports` repo, as tracking changes to `.gz` is cumbersome (see also 
- You should probably use release version (so, a tagged one) instead of commit hash
- if using release version, full source can be downloaded as a file (will probably be quicker than cloning the repo packages with/without submodules are available), for example:
  - full source with libs: https://github.com/micropython/micropython/releases/download/v1.15/micropython-1.15.tar.xz
  - only the main repo: https://github.com/micropython/micropython/archive/refs/tags/v1.15.tar.gz

pawelpisarczyk:
I agree with Marek. My other suggestion is to use unix directory instead of copying its content into phoenix directory. The main advantage of Phoenix-RTOS is the ability of providing the POSIX environment even on systems with small hardware resources, so it is not necessary to do specific port of Micropython.

niewim19:
Suggestions implemented: https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/9

niewim19:
phoenix-rtos/phoenix-rtos-ports#9 merged

Rearranged commits in this PR, so ready to go if no further issues are found.

nalajcie:
this might be my preference but I would make `ia32: ` prefix in commit message instead of `psh`/`ports` as the changes are related to `ia32` only, for example:
`ia32: ports: enable micropython building by default`
`ia32: psh: add pshlogin symlink`

No need to change it right now, just a remark for the future

_targets/build.project.ia32-generic
```diff
@@ -40,6 +41,7 @@ b_build_target() {
 	b_log "Building sample project for $TARGET"
 
 	# TODO: move to psh install target
+	ln -f "$PREFIX_ROOTFS/bin/"{psh,pshlogin}
```
nalajcie, line 44: this is unrelated change and not needed anymore, please remove and rebase on top of current master

nalajcie (changes requested):
I think we should enable micropython for `ia32-generic` and `armv7m7-imxrt106x` (and maybe `arm7va7-imx6ull` to verify against strange assumptions when using `arm-phoenix-` toolchain). 

We can do it even before the micropython PR would be merged - that way PRs in ports will try to build `micropython` (so these PRs would actually be checked).

Please update this PR so we can merge it.

PR #83, opened 2021-05-20
imxrt106x: update plo script

HBuczynski:


nalajcie:
Uch, I don't know why I should review this, maybe @gerard5 @jsarzy @Cinek28 could say more as they are actively using `plo` on rt1064?

nalajcie:
Ok, so that was a fix for submodule update? I hope we would introduce automatic testing for rt1064 soon, so it would get caught by CI.

You know, when You've added me as a reviewer, I assumed You want me to **review** the change :stuck_out_tongue:.

Probably adding one-sentence PR description would solve that :+1:. I think we should define PR process better (eg. some PRs are just for CI checks - no need to review them; some PRs were already tested internally, etc.), but that will take some time to polish.

HBuczynski:
You have right, I might have added some comments to make it more clear.

PR #84, opened 2021-05-24
project: define plo's preinit and user script in build.project*

HBuczynski:
This change introduces a new way of defining plo's script for `imxrt106x`, `imxrt117x` and `zynq-7000`.

There are two scripts:
- **preinit** - defines platform specific option, like: maps, phfs devices, console
- **user script** - defines user's program and  specific commands.

Currently, user's scripts contain basic application to run Phoenix-RTOS. However, user can add extra application to `PREINIT_SCRIPT` or `USER_SCRIPT`. When project should contain totally different setting, scripts can be completely overwritten. 

THIS PR is associated with changes in plo. Only after merging `feature/devs` to master, this PR can be merged.


_targets/build.project.armv7m7-imxrt117x
```diff
@@ -67,13 +114,15 @@ b_build_target() {
 	b_log "Building plo"
 	make -C plo clean all
 
-	cp "${PREFIX_PROG_STRIPPED}plo-ram-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.img" _boot/
-	cp "phoenix-rtos-hostutils/psu/plo-ram-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.sdp" _boot/
-
-	b_mkscript
+	MAGIC_USER_SCRIPT=$((($RANDOM % 0xffff) | (($RANDOM % 0xffff) << 16)))
```
agkaminski, line 117: I don't know, if I like MAGIC being random... I would rather have something that doesn't change between builds

agkaminski (commented):
Looks good, I see the point of making plo script dynamically. I'm no shell master, so I can't really review that

_targets/build.project.armv7a9-zynq7000
```diff
@@ -8,13 +8,32 @@
...
-	cat _targets/armv7a9-zynq7000.plo >> "$PREFIX_BUILD/plo/script.plo"
-	printf "\0" >> "$PREFIX_BUILD/plo/script.plo"
+b_build_target() {
+	b_log "Building sample project for $TARGET"
 
+	# FIXME: build plo out of tree and without clean
 	b_log "Building plo"
+	make -C plo clean all
+
+	b_mkscript_preinit
```
nalajcie, line 34: why can't we write script before `plo` and compile just once?

_targets/build.project.armv7a9-zynq7000
```diff
@@ -8,13 +8,32 @@
...
+
 
-	cat _targets/armv7a9-zynq7000.plo >> "$PREFIX_BUILD/plo/script.plo"
-	printf "\0" >> "$PREFIX_BUILD/plo/script.plo"
+b_build_target() {
+	b_log "Building sample project for $TARGET"
 
+	# FIXME: build plo out of tree and without clean
 	b_log "Building plo"
+	make -C plo clean all
```
nalajcie, line 32: why building with clean? will `plo` refactoring result in dropping `clean` from here?

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,24 +40,76 @@ KERNEL_OFFS=$((0x11000))
...
+
+				PROGS+=("$name")
+				psz=$((($(wc -c < "${PREFIX_PROG_STRIPPED}$name") + SIZE_PAGE - 1) & PAGE_MASK))
+				printf "alias %s 0x%x 0x%x\n" "$name" "$poffs" "$psz" >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
+				((poffs+=psz));;
+
+			kernel)
+				printf "alias %s 0x%x 0x%x\n" "phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf" "$KERNEL_OFFS" "$ksz" >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT";;
+		esac;
+		printf "%s\n" "$cmd" >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
```
nalajcie, line 93: instead of redirecting individual commands to file - redirect the whole loop (`for ... done >> file`)

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,24 +40,76 @@ KERNEL_OFFS=$((0x11000))
...
+
+b_mkscript_user() {
 	ksz=$((($(wc -c < "${PREFIX_PROG_STRIPPED}phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf") + SIZE_PAGE - 1) & PAGE_MASK))
 	poffs=$((KERNEL_OFFS + ksz))
 
-	for prog in "${PROGS[@]}"; do
-		psz=$((($(wc -c < "${PREFIX_PROG_STRIPPED}$prog") + SIZE_PAGE - 1) & PAGE_MASK))
-		printf "@%s(%x:%x)\n" "$prog" "$poffs" "$psz" >> "$PREFIX_BUILD/plo/script.plo"
-		((poffs+=psz))
+	printf "%08x" "$MAGIC_USER_SCRIPT" >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
```
nalajcie, line 73: can we have `\n` here - now  the file looks broken
image

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,24 +40,76 @@ KERNEL_OFFS=$((0x11000))
...
+	for cmd in "${USER_SCRIPT[@]}"; do
+		args=($cmd)
+		case "${args[0]}"
+		in
+			app)
+				if [ "${args[2]}" = "-x" ] ; then
+					name=$(cut -d ';' -f 1 <<< "${args[3]}")
```
nalajcie, line 80: why not `name="${args[3]}"`?

_targets/build.project.armv7m7-imxrt117x
nalajcie: - `$` not needed before `RANDOM` inside `$(())`
- I agree that building should be repeatable - building from the same tag/commit should always produce binary with the same crc32 (which is not the case in our buildsystem currently, but it might be desirable for MID certifications, etc.).

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,24 +40,76 @@ KERNEL_OFFS=$((0x11000))
...
-	cat _targets/armv7m7-imxrt106x.plo >> "$PREFIX_BUILD/plo/script.plo"
-	printf "\0" >> "$PREFIX_BUILD/plo/script.plo"
+	printf "\0" >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
 }
 
 
+b_mkscript_preinit() {
+	for cmd in "${PREINIT_SCRIPT[@]}"; do
+		printf "%s\n" "$cmd" >> "$PREFIX_BUILD/plo/script.plo"
+	done
```
nalajcie, line 103: one redirection outside of the loop?

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -67,7 +119,9 @@ b_build_target() {
 	b_log "Building plo"
 	make -C plo clean all
 
-	b_mkscript
+	MAGIC_USER_SCRIPT=$((($RANDOM % 0xffff) | (($RANDOM % 0xffff) << 16)))
+	b_mkscript_user
```
nalajcie, line 123: I don't like that the scripts are implicitly removed with each recompilation by `make -C plo clean`. As they are created by the script, they should be "managed" by the script (so - explicitly removed by it when needed?)

_targets/build.project.armv7m7-imxrt117x
```diff
@@ -40,24 +40,71 @@ KERNEL_OFFS=$((0x11000))
...
+                "wait 2000")
+
+# Obligatory applications to run Phoenix-RTOS
+USER_SCRIPT=("kernel ${BOOT_DEVICE}"
+             "app ${BOOT_DEVICE} -x dummyfs xip1 ocram2"
+             "app ${BOOT_DEVICE} -x imxrt-multi ocram2 ocram2"
+             "app ${BOOT_DEVICE} -x psh xip1 ocram2")
+
+
+b_mkscript_user() {
```
nalajcie, line 65: code duplication with rt1064

nalajcie (commented):
pointed out some of the things I don't like, not all of them need to be addressed right away ;)

_targets/build.project.armv7m7-imxrt117x
HBuczynski: After building plo, there are two images: 
- `plo_ram.img` - images is loaded to ram and is used to flash phoenix.disk
- `plo.img` - image which is a part of phoenix.disk

Both of the images contain preinit script in `.data` section with command: `call <dev> <script name> <magic>`. If you flash your memory for the first time, it's ok, commad `call` is not executed because `MAGIC` is incorrect. However, if you want to re-flash memory with the `MAGIC` defines as constant, execution of `call` will be fine. The side effect will be loading kernel to ram, where is currently plo.

I have discussed this approach with @pawelpisarczyk. The other way is to create the separate script for `plo_ram.img` without `call` command.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Application can be defined in two ways:
- `app ${BOOT_DEVICE} -x psh xip1 ocram2` - name is located on 3rd position
- `app ${BOOT_DEVICE} imxrt-flash ocram2 ocram2` - name is located on 2nd position

_targets/build.project.armv7m7-imxrt117x
HBuczynski: Yeah, I know. 
@pawelpisarczyk asked to not create additional files if it is not necessary so I try to avoid such situation. Currently it is only one function. @pawelpisarczyk what do you think abut creating common file? Currently it is only used by `imxrt` family but it can change in the future and can be used by other platforms.

_targets/build.project.armv7m7-imxrt117x
agkaminski: Wouldn't it be better to have separate script for plo_ram.img? It is used only to program img after all

_targets/build.project.armv7a9-zynq7000
HBuczynski: Removed. There is only one compilation.

_targets/build.project.armv7a9-zynq7000
HBuczynski: Removed. There is only one compilation without `clean`.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Done.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Compilation with clean is removed. Script checks whether file exists and removes it or not.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Thanks for notice. Done.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Yeah, it can be changed. 

However I would like to do it in separate PR because this change will affect plo which is currently in a merging process.

_targets/build.project.armv7a9-zynq7000
```diff
@@ -8,13 +8,33 @@
...
+                "console 0.1"
+                "wait 1500")
+
+
+b_mkscript_preinit() {
+	if [ -f "$PREFIX_BUILD/plo/script.plo" ]; then
+		rm "$PREFIX_BUILD/plo/script.plo"
+	else
+		mkdir -p "$PREFIX_BUILD/plo/"
+	fi
```
kemonats, line 25:
```suggestion
	mkdir -p "$PREFIX_BUILD/plo/"
	rm -f "$PREFIX_BUILD/plo/script.plo"
```
without `if`

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,36 +40,99 @@ KERNEL_OFFS=$((0x11000))
...
+             "app ${BOOT_DEVICE} imxrt-flash ocram2 ocram2")
+
+
+b_mkscript_user() {
+
+	if [ -f "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT" ]; then
+		rm "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
+	else
+		mkdir -p "$PREFIX_BUILD/plo/"
+	fi
```
kemonats, line 75:
```suggestion
	mkdir -p "$PREFIX_BUILD/plo/"
	rm -f "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
```
without `if`

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,36 +40,99 @@ KERNEL_OFFS=$((0x11000))
...
+		for cmd in "${USER_SCRIPT[@]}"; do
+			args=($cmd)
+			case "${args[0]}"
+			in
+				app)
+					if [ "${args[2]}" = "-x" ] ; then
+						name=$(cut -d ';' -f 1 <<< "${args[3]}")
+					else
+						name=$(cut -d ';' -f 1 <<< "${args[2]}")
+					fi
```
kemonats, line 91:
```suggestion
					if [ "${args[2]}" = "-x" ] ; then
						name=$(${args[3]}%%;*)
					else
						name=$(${args[2]}%%;*)
					fi
```
It would be better to use parameter substitution.

_targets/build.project.armv7m7-imxrt106x
```diff
@@ -40,36 +40,99 @@ KERNEL_OFFS=$((0x11000))
...
+		printf "\0"
+	} >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
 }
 
 
+b_mkscript_preinit() {
+	file="$PREFIX_BUILD/plo/script.plo"
+	if [ -f "$file" ]; then
+		rm "$file"
+	fi
```
kemonats, line 112:
```suggestion
	rm -f "$file"
```
without `if`

_targets/build.project.armv7m7-imxrt117x
```diff
@@ -40,40 +40,100 @@ KERNEL_OFFS=$((0x11000))
...
+
+	if [ -f "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT" ]; then
+		rm "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
+	else
+		mkdir -p "$PREFIX_BUILD/plo/"
+	fi
```
kemonats, line 71:
```suggestion
	mkdir -p "$PREFIX_BUILD/plo/"
	rm -f "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
```
without `if`

_targets/build.project.armv7m7-imxrt117x
```diff
@@ -40,40 +40,100 @@ KERNEL_OFFS=$((0x11000))
...
+		for cmd in "${USER_SCRIPT[@]}"; do
+			args=($cmd)
+			case "${args[0]}"
+			in
+				app)
+					if [ "${args[2]}" = "-x" ] ; then
+						name=$(cut -d ';' -f 1 <<< "${args[3]}")
+					else
+						name=$(cut -d ';' -f 1 <<< "${args[2]}")
+					fi
```
kemonats, line 87:
```suggestion
					if [ "${args[2]}" = "-x" ] ; then
						name=$(${args[3]}%%;*)
					else
						name=$(${args[2]}%%;*)
					fi
```
It would be better to use parameter substitution.

_targets/build.project.armv7m7-imxrt117x
```diff
@@ -40,40 +40,100 @@ KERNEL_OFFS=$((0x11000))
...
+		printf "\0"
+	} >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
 }
 
 
+b_mkscript_preinit() {
+	file="$PREFIX_BUILD/plo/script.plo"
+	if [ -f "$file" ]; then
+		rm "$file"
+	fi
```
kemonats, line 108:
```suggestion
	rm -f "$file"
```
without `if`

_targets/build.project.armv7m7-imxrt117x
HBuczynski: Yeah, it looks better, thanks. Done.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Yeah, it looks better, thanks. Done.

_targets/build.project.armv7m7-imxrt117x
HBuczynski: Yeah, it looks better, thanks. Done.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Yeah, it looks better, thanks. Done.

_targets/build.project.armv7a9-zynq7000
HBuczynski: Yeah, it looks better, thanks. Done.

_targets/build.project.armv7m7-imxrt117x
HBuczynski: Yeah, shorter approach is better. Thanks.
However, you have made a small mistake, there should be `${args[3]%%;*}` instead of `${${args[3]}%%;*}`.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Yeah, shorter approach is better. Thanks.
However, you have made a small mistake, there should be `${args[3]%%;*}` instead of `${${args[3]}%%;*}`.

agkaminski:
Script for imxrt117x is invalid - addresses doesn't have `0x` prefix, there is no interface for phoenixd defined.

_targets/build.project.armv7m7-imxrt117x
agkaminski: I'm getting` Magic number for user.plo is wrong.` when loading plo to the RAM multiple times, but it doesn't really change anything...

agkaminski (approved):
I'm still not a fan of `MAGIC` being generated via `RANDOM`, otherwise LGTM

_targets/build.project.armv7m7-imxrt106x
nalajcie: I hoped that empty lines are allowed in plo script, so adding `\n` here would actually make first line of the script empty and be transparent for the parser

_targets/build.project.armv7m7-imxrt106x
nalajcie: yeah, I was asking about getting ";" out of the name - I suppose it has something to do with arguments to the program. I suppose the plo script syntax should be documented somewhere.

Resolving as not relevant anymore.

_targets/build.project.armv7m7-imxrt106x
HBuczynski: Yes, you have right. The new line has been added after `MAGIC NUMBER`.

HBuczynski:
This PR contains update of the plo to the newest version, so that it can be merged.

There is only remaining issue about `MAGIC` being generated via `RANDOM`.

_targets/build.project.armv7m7-imxrt117x
HBuczynski: It is expected behavior. If You don't have random `MAGIC`, the command `call` will start execution of script located in flash. After that the kernel will be loaded to `ITCM` and plo will be overwritten and the application will crash.
So that I think that using random number in this case changes something ;)

pawelpisarczyk:
Let's merge. In my opinion MAGIC allows to solve many problems (upgrade etc). If it will be problematic then we will remove it.

PR #89, opened 2021-05-27
project: add example of user app

HBuczynski:
The main goal of this PR is to simplify usage of phoenix-rtos-project.
It contains the following changes:

1. **_user** - directory contains user applications and is built for each platform; output binaries are located in _/root/userbin_
2. For `imxrt106x`, `imxrt117x` and `zynq-7000`, programs are added to plo's script only if variable `TARGET_PROJECT` is empty.
3. One option is to add building of **_user** to each of the scripts in **phoenix-rtos-build/build-core-***. In this scenario **_user** is built each time.
4. The other approach is to create **_project** directory contains files for each for the platform, like `build.project-*-*-user` and there would be located building of **_user**. In this case, **_user** directory is built only for specific target.

**BEFORE MERGE:** Options 3 or 4 have to be chosen and files `_targets/build.project.*` have to be updated for each of the platform.

### DIFF
--- a/_targets/build.project.armv7m7-imxrt106x
+++ b/_targets/build.project.armv7m7-imxrt106x
@@ -96,8 +96,10 @@ b_mkscript_user() {
 			esac;
 			printf "%s\n" "$cmd"
 		done
+
 		printf "\0"
 	} >> "$PREFIX_BUILD/plo/$NAME_USER_SCRIPT"
+
 }
 
 
@@ -116,6 +118,12 @@ b_mkscript_preinit() {
 		printf "call %s %s %08x\n" "$BOOT_DEVICE" "$NAME_USER_SCRIPT" "$MAGIC_USER_SCRIPT"
 		printf "\0"
 	} >> "$file"
+}
+
+
+b_build_project() {
+	b_log "Building user applications"
+	make -C "_user" all install
 }
 
 
--- a/_user/Makefile
+++ b/_user/Makefile
@@ -1,0 +1,26 @@
+#
+# Makefile for user app
+#
+# Copyright 2021 Phoenix Systems
+#
+# %LICENSE%
+#
+
+include ../phoenix-rtos-build/Makefile.common
+
+.DEFAULT_GOAL := all
+
+# default path for the programs to be installed in rootfs
+DEFAULT_INSTALL_PATH := /usr/bin
+
+# read out all components
+ALL_MAKES := $(shell find . -mindepth 2 -name Makefile)
+include $(ALL_MAKES)
+
+# create generic targets
+DEFAULT_COMPONENTS := $(ALL_COMPONENTS)
+
+.PHONY: all install clean
+all: $(DEFAULT_COMPONENTS)
+install: $(patsubst %,%-install,$(DEFAULT_COMPONENTS))
+clean: $(patsubst %,%-clean,$(ALL_COMPONENTS))
--- a/_user/hello/main.c
+++ b/_user/hello/main.c
@@ -1,0 +1,24 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Hello World
+ *
+ * Example of user application
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+
+
+int main(void)
+{
+	printf("Hello World!!\n");
+
+	return 0;
+}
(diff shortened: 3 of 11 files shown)


_targets/build.project.armv7m7-imxrt106x
```diff
@@ -96,8 +96,21 @@ b_mkscript_user() {
 			esac;
 			printf "%s\n" "$cmd"
 		done
+
+		if [[ -z "${TARGET_PROJECT}" ]] && [[ -d "${PREFIX_ROOTFS}"userbin ]]; then
```
nalajcie, line 100: `[` is enough

_user/Makefile
```diff
@@ -0,0 +1,28 @@
...
+# Makefile for user app
+#
+# Copyright 2021 Phoenix Systems
+#
+# %LICENSE%
+#
+
+include ../phoenix-rtos-build/Makefile.common
+
+CFLAGS += $(BOARD_CONFIG)
```
nalajcie, line 11: board config should not be needed for userspace apps

_user/Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+.DEFAULT_GOAL := all
+
+DEFAULT_COMPONENTS := hello
+
+# default path for the programs to be installed in rootfs
+DEFAULT_INSTALL_PATH := /userbin
```
nalajcie, line 18: why not `/usr/bin` ?

_user/Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+# read out all components
+ALL_MAKES := $(wildcard */Makefile) $(wildcard */*/Makefile) $(wildcard */*/*/Makefile)
```
nalajcie, line 21: too many wildcards, safe to use only the first one or `find` approach from `phoenix-rtos-test`

_user/hello/main.c
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+
+
+int main(void) {
```
nalajcie, line 19: coding style error

nalajcie (commented):
My remarks:
 - maybe we can add `b_build_project` function in `_target/*` which would compile the examples? (and then overwrite the function in custom project in `_project/*` not to build the examples)
 - I would probably create "groups" of examples (`Makefile` phony targets, eg `text`, `tcpip`, `graph`, etc.) in _user to explicitly state which example groups should be built for each target
 - **I would go for adding user apps to rootfs and focus on bringing rootfs to other platforms (`riscv`, `rt106x`) instead of writing custom plo scripts for uploading them** (this will break soon as we may have too many sample apps to fit at once :)). I'm more than OK with bringing the samples only to platforms with rootfs with this PR and work on rootfs for other platforms in next PRs.

pawelpisarczyk:
Please remove TARGET_PROJECT variable. We have to many variables parametrizing something - this is bad approach. 

Programs from the _user directory should be built each time and this building should be invoked from _target/build.project.<TARGET> script in b_build_target() function. Programs should be installed in /usr/bin or /usr/sbin directories depending on the type of user binary.

Currently, the running of such user binaries is unavailable on platforms without MMU because on these platforms support of memory objects is trivialized and will be added in the near future.

_user/Makefile
pawelpisarczyk: We agreed on /usr/bin /usr/sbin. I don't know why.

_targets/build.project.armv7m7-imxrt106x
pawelpisarczyk: I don't understand the reason of introducing this variable. We have a bunch of redundant variables, defines and types. Many of them are used only one time.

_user/Makefile
pawelpisarczyk: I agree with  Marek.

_user/Makefile
pawelpisarczyk: Marek, it seems that you read in my minds. The only difference is that I recommend to use `b_build_target` function instead of defining new build function.

nalajcie: My reasoning was that `b_build_project` is already invoked in generic `build.project`:

https://github.com/phoenix-rtos/phoenix-rtos-project/blob/248c62ecec7467c849f91e2b2204b6de54638193/build.project#L62-L65

And it's easy to overwrite/undefine full function in custom project not to build sample binaries (just define Your own in `_projects/build.project.ia32-generic-myproject`).

If the build would be part of `b_build_target` it would probably be harder to disable them in custom projects :).

nalajcie (approved):
LGTM.

PR #91, opened 2021-05-28
psh: add defuser hash in _targets

niewim19:
defuser password hash is added to each target using psh for testing of login feature in psh.
Hash corresponds to password: 1234

JIRA: PD-75


nalajcie (approved):
Looks good, some remarks:
 - I was eager to propose syntax `: "${PSH_DEFUSRPWDHASH=asdf}"` to be able to get the variable from ENV, but I suppose nobody sane would like to set password **hash** from the env directly
 - I think `psh` is not used on stm32, but I see it's being compiled for that target so adding the define there is valid for now ;).

PR #99, opened 2021-06-11
_user/rotrectangle: added user example - rotating rectangle program - draft

damianloew:
JIRA: PD-119

### DIFF
--- a/_user/rotrectangle/Makefile
+++ b/_user/rotrectangle/Makefile
@@ -8,5 +8,6 @@ NAME := rotrectangle
 LOCAL_SRCS := main.c
 LIBS := libgraph libvga libvirtio
 
-include $(binary.mk)
-
+ifneq ($(TARGET_FAMILY), host-pc)
+	include $(binary.mk)
+endif
--- a/_user/rotrectangle/Makefile
+++ b/_user/rotrectangle/Makefile
@@ -1,0 +1,12 @@
+#
+# Makefile for user application
+#
+# Copyright 2021 Phoenix Systems
+#
+
+NAME := rotrectangle
+LOCAL_SRCS := main.c
+LIBS := libgraph libvga libvirtio
+
+include $(binary.mk)
+
--- a/_user/rotrectangle/main.c
+++ b/_user/rotrectangle/main.c
@@ -1,0 +1,273 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Rotating Rectangle
+ *
+ * Example of user application
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Damian Loewnau
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+
+/*					a
+*	(x0,y0)	.................
+*			|				|
+*			|		.		|
+*		b	|	(xc,yc)		|
+*			|...............|
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <graph.h>
+#include <time.h>
+#include <unistd.h>
+#include <math.h>
+#include <termios.h>
+
+#include <sys/threads.h>
+
+typedef struct {
+	unsigned int xc;
+	unsigned int yc;
+	unsigned int a;
+	unsigned int b;
+	unsigned int color;
+} rectangle_t;
+
+typedef struct {
+	rectangle_t rec;
+	rectangle_t toRemove;
+	unsigned int rotSpeed;
+	unsigned int maxxc;
+	unsigned int maxyc;
+	unsigned int minxc;
+	unsigned int minyc;
+	char stop;
+} rotrectangle_t;
+
+/* Enables/disables canon mode and echo */
+static int rotrectangle_switchmode(int canon)
+{
+	struct termios state;
+
+	tcgetattr(STDIN_FILENO, &state);
+	if (canon) {
+		state.c_lflag |= ICANON;
+		state.c_lflag |= ECHO;
+	} else {
+		state.c_lflag &= ~ICANON;
+		state.c_lflag &= ~ECHO;
+		state.c_cc[VMIN] = 1;
+	}
+	tcsetattr(STDIN_FILENO, TCSANOW, &state);
+	
+	return 0;
+}
+
+
+static void rotrectangle_getkey(void *arg)
+{
+	char ch;
+	rotrectangle_t *self = (rotrectangle_t *)arg;
+	while ((self->stop == 0) && (ch = getchar())) {
+		if (ch == 27) {					/*ESC code from arrow keys*/
+			getchar();
+			ch = getchar();
+		}
+		switch (ch) {
+			case 67:
+				if (self->rec.xc < (self->maxxc-2))
+					self->rec.xc+=3;
+				break;
+			case 68:
+				if (self->rec.xc > (self->minxc+2))
+					self->rec.xc-=3;
+				break;
+			case 65:
+				if (self->rec.yc > (self->minyc+2))
+					self->rec.yc-=3;
+				break;
+			case 66:
+				if (self->rec.yc < (self->maxyc-2))
+					self->rec.yc+=3;
+				break;
+			case 43:
+				if (self->rotSpeed < 5)
+					self->rotSpeed++;
+				break;
+			case 45:
+				if (self->rotSpeed > 1)
+					self->rotSpeed--;
+				break;
+			case 'r':
+				self->rec.color = 4;
+				break;
+			case 'g':
+				self->rec.color = 2;
+				break;
+			case 'b':
+				self->rec.color = 1;
+				break;
+			case 'q':
+				self->stop = 1;
+				break;
+			default:
+				break;
+		}
+	}
+}
+
+static int rotrectangle_print(rectangle_t *self, graph_t *graphp, unsigned int angle)
+{
+	double dx, dy, angleRad, xcDouble, ycDouble, aDouble, bDouble;
+	unsigned int x0, y0, x, y;		/* (x,y) is start point for next rectangle lines */
+
+	xcDouble = (double)self->xc;
+	ycDouble = (double)self->yc;
+	aDouble = (double)self->a;
+	bDouble = (double)self->b;
+
+	angleRad = ((double)(angle)*3.141592)/180.0;
+	x0 = (unsigned int)(xcDouble - ( (sin(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
+	y0 = (unsigned int)(ycDouble - ( (cos(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
+
+	x = x0;
+	y = y0;
+	dx = (self->a)*cos(angleRad); 
+	dy = (self->a)*sin(angleRad);
+	graph_line(graphp, x, y, (int)dx, (int)dy, 1, self->color, GRAPH_QUEUE_HIGH);
+
+	x = x0+dx;
+	y = y0+dy;
+	dx = (self->b)*cos(angleRad+(0.5*3.141592));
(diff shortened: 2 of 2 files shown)


pawelpisarczyk:
Damian, please correct coding style

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+{
+	double dx, dy, angleRad, xcDouble, ycDouble, aDouble, bDouble;
+	unsigned int x0, y0, x, y;		/* (x,y) is start point for next rectangle lines */
+
+	xcDouble = (double)self->xc;
+	ycDouble = (double)self->yc;
+	aDouble = (double)self->a;
+	bDouble = (double)self->b;
+
+	angleRad = ((double)(angle)*3.141592)/180.0;
```
agkaminski, line 136: Use `M_PI` instead of typing pi value by yourself

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+	x0 = (unsigned int)(xcDouble - ( (sin(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
+	y0 = (unsigned int)(ycDouble - ( (cos(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
+
+	x = x0;
+	y = y0;
+	dx = (self->a)*cos(angleRad); 
+	dy = (self->a)*sin(angleRad);
```
agkaminski, line 143: There should be space between `*`, `+`, etc. operands, e.g.:
`dx = (self->a) * cos(angleRad);`

Applies to other occurrences

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+	graph_line(graphp, x, y, (int)dx, (int)dy, 1, self->color, GRAPH_QUEUE_HIGH);
+
+	x = x0+dx;
+	y = y0+dy;
+	dx = (self->b)*cos(angleRad+(0.5*3.141592));
+	dy = (self->b)*sin(angleRad+(0.5*3.141592));
```
agkaminski, line 149: Use `M_PI_2`

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+	graph_line(graphp, x, y, (int)dx, (int)dy, 1, self->color, GRAPH_QUEUE_HIGH);
+
+	x = x0;
+	y = y0;
+	dx = (self->b)*cos(angleRad+(0.5*3.141592)); 
+	dy = (self->b)*sin(angleRad+(0.5*3.141592));
```
agkaminski, line 155: Use `M_PI_2`

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+	unsigned int x0, y0, x, y;		/* (x,y) is start point for next rectangle lines */
+
+	xcDouble = (double)self->xc;
+	ycDouble = (double)self->yc;
+	aDouble = (double)self->a;
+	bDouble = (double)self->b;
+
+	angleRad = ((double)(angle)*3.141592)/180.0;
+	x0 = (unsigned int)(xcDouble - ( (sin(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
+	y0 = (unsigned int)(ycDouble - ( (cos(M_PI/2-angleRad-atan(bDouble/aDouble)))*sqrt(pow(aDouble, 2.0)+pow(bDouble, 2.0)) )/2.0);
```
agkaminski, line 138: Use `M_PI_2`

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+static int rotrectangle_rotating(rotrectangle_t *self, graph_t *graphp)
+{
+	unsigned int alfa, ret;
+	ret = 0;
+	while (self->stop == 0) {
+		for (alfa=0; alfa<360; alfa+=self->rotSpeed) {
+			if (self->stop != 0)
+				break;
+			ret = rotrectangle_print(&(self->rec), graphp, alfa);
+			// graph_fill(graphp, self->rec.xc, self->rec.yc, self->rec.color, GRAPH_FILL_FLOOD, GRAPH_QUEUE_HIGH);
```
agkaminski, line 187: Why is this line commented out?

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,273 @@
...
+	printf("  Press any key to start: \n");
+	getchar();
+
+	return 0;
+}
+
+int main(void)
+{
+	int ret;
+	char stack[1024] __attribute__ ((aligned(8)));
```
agkaminski, line 221: It's better to put big memory objects in .bss instead of stack.

agkaminski (changes requested):
View files changed to see review about coding style from reviewdog.
Design LGTM, but style fixes are needed

_user/rotrectangle/main.c
damianloew: There is a problem with screen refreshing when I use filled rectangle (refresh occurs after displaying frame and before displaying filling) and Łukasz said I can comment it out now. So this is a demo without it.

agkaminski: I think a comment about it would be helpful, so we won't forget why it is commented out for now

_user/rotrectangle/main.c
pawelpisarczyk: Please don't use semi-hungarian convention in naming variables. `aDouble` and `bDouble` variables should be named just 'a' and 'b'. Types of them don't matter.

_user/rotrectangle/main.c
pawelpisarczyk: Please take a look on ma previous comment related to variable naming.

_user/rotrectangle/main.c
pawelpisarczyk: Please comment this line using C comments  ( /* */ ) instead of C++ double /. Rotating empty rectangle is pretty enough for purposes of this demo.

_user/rotrectangle/main.c
pawelpisarczyk: This is general remark. It is even better to allocate thread stack using `malloc()`.

nalajcie:
Just devel process note - if You want to convert PR to draft, just click "Convert to draft" below the "Reviewers" (don't edit the PR title):
image

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+	tcsetattr(STDIN_FILENO, TCSANOW, &state);
+
+	return 0;
+}
+
+
+static void rotrectangle_getkey(void *arg)
+{
+	char ch;
+	rotrectangle_t *self = (rotrectangle_t *)arg;
```
nalajcie, line 81: Error: writing/reading the same struct from 2 threads which is not marked as `volatile` - may work just by accident. Please fix it.

There still might be glitches due to not guarding better the access to the struct, but from the current code I assume they would be minimal.

_user/rotrectangle/Makefile
```diff
@@ -0,0 +1,13 @@
...
+#
+# Copyright 2021 Phoenix Systems
+#
+
+NAME := rotrectangle
+LOCAL_SRCS := main.c
+LIBS := libgraph libvga libvirtio
+
+ifneq ($(TARGET_FAMILY), host-pc)
+	include $(binary.mk)
```
nalajcie, line 12: use spaces for indentation in Makefiles outside of receipes.

I don't think `_user` folder would ever be built for `host-pc` target (if it somehow is - it's a build error and should be fixed elsewhere), so we don't need to guard against it.

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+	rotrectangle.rec.a = 180;
+	rotrectangle.rec.b = 90;
+	rotrectangle.rec.color = 60000;
+
+	ret = rotrectangle_startpanel(&rotrectangle);
+
+	rotrectangle_switchmode(0);
+
+	if ((ret = graph_init()) < 0) {
+		fprintf(stderr, "test_graph: failed to initialize library\n");
```
nalajcie, line 243: not a big mistake, but leaking `stack` (will be cleared by the OS, but valgrind would show memory leak nevertheless)

Applies to other early-returns

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+		return ret;
```
nalajcie, line 244: do not return negative value from the `main` as it's not well-defined. Stick to 0-255 values (the only available to be read by the parent process by `waitpid`). Not sure if POSIX requirement, but on Linux exit code >127 means the program was killed with singal (and the signal number would be `$? - 127`) so it's even better to return value from the range 0-127.

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+			/* filling rectangle functions are commented out, because of problem with screen refreshing, it's demo without it */
+			/* graph_fill(graphp, self->rec.xc, self->rec.yc, self->rec.color, GRAPH_FILL_FLOOD, GRAPH_QUEUE_HIGH); */
+			ret = rotrectangle_save(self);
+			usleep(10000); /* min value */
+			ret = rotrectangle_print(&(self->prev), graphp, alfa);
+			/* graph_fill(graphp, self->rec.xc, self->rec.yc, 0, GRAPH_FILL_FLOOD, GRAPH_QUEUE_HIGH); */
+		}
+	}
+
+	return ret;
```
nalajcie, line 196: not sure why `ret` is needed as it's never checked and all of the functions return 0

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+
+	rotrectangle.speed = 3;
+	rotrectangle.stop = 0;
+	rotrectangle.rec.a = 180;
+	rotrectangle.rec.b = 90;
+	rotrectangle.rec.color = 60000;
+
+	ret = rotrectangle_startpanel(&rotrectangle);
+
+	rotrectangle_switchmode(0);
```
nalajcie, line 240: maybe switch mode only after all initialization has succeeded?

Aaand - it would be good to switch it back to the previous mode at exit (in case shell won't do it for You)

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+
+	rotrectangle_t rotrectangle;
+
+	rotrectangle.speed = 3;
+	rotrectangle.stop = 0;
+	rotrectangle.rec.a = 180;
+	rotrectangle.rec.b = 90;
+	rotrectangle.rec.color = 60000;
+
+	ret = rotrectangle_startpanel(&rotrectangle);
```
nalajcie, line 238: `ret` not checked and overwritten 2 lines below, no need to set it here (dead-store)

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+			default:
+				break;
+		}
+	}
+
+	endthread();
+}
+
+
+static int rotrectangle_print(rectangle_t *self, graph_t *graphp, unsigned int angle)
```
nalajcie, line 124: could this be `const rectangle_t *self` ?

`static void`

_user/rotrectangle/main.c
```diff
@@ -0,0 +1,277 @@
...
+	y = y0 + dy;
+	dx = (self->a) * cos(alfa);
+	dy = (self->a) * sin(alfa);
+	graph_line(graphp, x, y, (int)dx, (int)dy, 1, self->color, GRAPH_QUEUE_HIGH);
+
+	return 0;
+}
+
+
+static int rotrectangle_save(rotrectangle_t *self)
```
nalajcie, line 166: `static void`

nalajcie (commented):
There are things which need need to be fixed as a separate PR as the original was merged too soon.

Please fix invalid stack size for the thread ASAP.

PR #104, opened 2021-06-17
_user/rotrectangle: correct rotating rectangle program

damianloew:
Correct rotating rectangle program

## Description

Except for minor fixes:
Stack size for thread has been changed
Mutex functions has been added for better synchronization
Volatile prefix has been added for variables used in thread

JIRA: PD-119

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic - qemu).

## Checklist:

## Special treatment

### DIFF
--- a/_user/rotrectangle/main.c
+++ b/_user/rotrectangle/main.c
@@ -34,8 +34,8 @@
 
 
 typedef struct {
-	unsigned int xc;
-	unsigned int yc;
+	volatile unsigned int xc;
+	volatile unsigned int yc;
 	unsigned int a;
 	unsigned int b;
 	unsigned int color;
@@ -43,9 +43,10 @@ typedef struct {
 
 
 typedef struct {
+	handle_t lock;
 	rectangle_t rec;
 	rectangle_t prev;
-	unsigned int speed;
+	volatile unsigned int speed;
 	unsigned int maxxc;
 	unsigned int maxyc;
 	unsigned int minxc;
@@ -55,7 +56,7 @@ typedef struct {
 
 
 /* Enables/disables canon mode and echo */
-static int rotrectangle_switchmode(int canon)
+static void rotrectangle_switchmode(int canon)
 {
 	struct termios state;
 
@@ -70,8 +71,6 @@ static int rotrectangle_switchmode(int canon)
 		state.c_cc[VMIN] = 1;
 	}
 	tcsetattr(STDIN_FILENO, TCSANOW, &state);
-
-	return 0;
 }
 
 
@@ -84,6 +83,7 @@ static void rotrectangle_getkey(void *arg)
 			getchar();
 			ch = getchar();
 		}
+		mutexLock(self->lock);
 		switch (ch) {
 			case 67:
 				if (self->rec.xc < (self->maxxc - 4))
@@ -115,13 +115,14 @@ static void rotrectangle_getkey(void *arg)
 			default:
 				break;
 		}
+		mutexUnlock(self->lock);
 	}
 
 	endthread();
 }
 
 
-static int rotrectangle_print(rectangle_t *self, graph_t *graphp, unsigned int angle)
+static void rotrectangle_print(const rectangle_t *self, graph_t *graphp, unsigned int angle)
 {
 	double dx, dy, alfa, xc, yc, a, b;
 	unsigned int x0, y0, x, y; /* (x,y) is start point for next rectangle lines */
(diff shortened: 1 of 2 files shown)


_user/rotrectangle/main.c
```diff
@@ -210,32 +201,34 @@ static int rotrectangle_startpanel(rotrectangle_t *self)
...
 	getchar();
-
-	return 0;
 }
 
 
 int main(void)
 {
 	int ret;
+	unsigned int stacksz;
```
agkaminski, line 210: `static const unsigned int stacksz = 1024;`

_user/rotrectangle/main.c
```diff
@@ -210,32 +201,34 @@ static int rotrectangle_startpanel(rotrectangle_t *self)
...
 	graph_adapter_t adapter = GRAPH_ANY;
 	graph_mode_t mode = GRAPH_DEFMODE;
 	graph_freq_t freq = GRAPH_DEFFREQ;
 	graph_t graph;
 
 	rotrectangle_t rotrectangle;
 
+	stacksz = 1024;
+
+	if ((stack = (char *)malloc(stacksz)) == NULL)
```
agkaminski, line 222: Casting malloc result is a bad practice:

https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc

_user/rotrectangle/main.c
```diff
@@ -137,67 +135,60 @@ static int rotrectangle_print(rectangle_t *self, graph_t *graphp, unsigned int a
...
 	self->prev.a = self->rec.a;
 	self->prev.b = self->rec.b;
 	self->prev.color = 0;
-
-	return 0;
 }
 
 
-static int rotrectangle_rotating(rotrectangle_t *self, graph_t *graphp)
+static void rotrectangle_rotating(rotrectangle_t *self, graph_t *graphp)
```
agkaminski, line 172: Now `ret` is unused. Remove it

damianloew: ret is still used in libgraph init

agkaminski: GitHub Actions
/ build image (armv7a7-imx6ull)

_user/rotrectangle/main.c#L174

unused variable 'ret' [-Wunused-variable]

_user/rotrectangle/main.c
pawelpisarczyk: Alek, I don't understand your remark. This line is correct and such construction is widely used in Phoenix-RTOS code.

nalajcie:
notes regarding the process:
 - Your PR is a bugfix (not a breaking change, "breaking change" is when other components depend on Your component and You're introducing changes which will break other components without additional fixes in them)
 - Please update the PR checklist accordingly - especially the two bottom ones (no linting errors and introducing no new warnings - **which You have failed to do** - as Alek pointed out)

_user/rotrectangle/main.c
agkaminski: There is no need to cast explicitly `void *` to any other type. Some say it's a bad practice, other say otherwise. If you think this is ok, I won't oppose.

agkaminski:
Please use squash option when merging this PR

_user/rotrectangle/main.c
nalajcie: https://github.com/phoenix-rtos/phoenix-rtos-project/runs/2849294798?check_suite_focus=true#step:4:660

nalajcie (changes requested):
Please note that some of the requested changes from the merged PR are not yet implemented: https://github.com/phoenix-rtos/phoenix-rtos-project/pull/99

_user/rotrectangle/main.c
```diff
@@ -220,24 +209,22 @@ int main(void)
...
 
-	stack = (char *)malloc(1024 * sizeof(char));
-
 	graph_adapter_t adapter = GRAPH_ANY;
 	graph_mode_t mode = GRAPH_DEFMODE;
 	graph_freq_t freq = GRAPH_DEFFREQ;
 	graph_t graph;
 
-	rotrectangle_t rotrectangle;
+	volatile rotrectangle_t rotrectangle;
```
agkaminski, line 217: Honestly, I don't see the point of passing this structure around. Wouldn't it be much simpler to declare it as a global variable?

pawelpisarczyk: Alek, the idea of passing structure is better then define it globally because you can easily extend this program with threads rotating many rectangles on the same screen

agkaminski: Ok, I was actually wondering if this is a motivation behind this. So only fix I see here is to make only some members of `rotrectangle_t` `volatile`, not whole `struct`.

agkaminski (approved):
LGTM

_user/rotrectangle/main.c
nalajcie: Oh, I thought we will go with lock-less programming, thus suggested `volatile`.

I'm not sure if our mutex'es are written correctly (if acquire/release operations include necessary memory barriers), but with them `volatile` should not be needed (see explanations in linux kernel docs: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/volatile-considered-harmful.rst).

Please note that You missed `volatile` on `stop` which is shared between the threads.

Side note: I'm only reading the code, but I wonder if it's possible to slow down the render thread by writing may chars at once, maybe input handling thread should have lower priority?

_user/rotrectangle/main.c
```diff
@@ -264,14 +256,23 @@ int main(void)
...
+	beginthread(rotrectangle_getkey, 4, stack, stacksz, (void *)&rotrectangle);
+
+	rotrectangle_rotating(&rotrectangle, &graph);
+
+	rotrectangle_switchmode(1);
 
 	graph_close(&graph);
 	graph_done();
 
 	free(stack);
```
nalajcie, line 275: I don't know if we have `join` in our native api, but it could be that we would free the stack of still-running thread

nalajcie (approved):
Please note that I haven't run the code, so it's only static analysis on my part.

The fixes improve the currrent code considerably - let's merge them. :+1: 

I've left some comments for consideration in the future.

damianloew:
Thanks for any comments, I will apply the mentioned suggestions in the future

PR #110, opened 2021-06-22
_user/voxeldemo: Add double buffering and make run on "any" graph video adapter

gerard5:
RTOS: 57

## Description

This PR adds double buffering to fix fickering on Cirrus adapter and enables to run demo on "any" graph video adapter which support 32bit true color.

## Motivation and Context

Fixes flickering on cirrus.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32 with and without kvm hwaccel, using virtiogpu as well cirrus qemu adapter).

## Checklist:

## Special treatment

### DIFF
--- a/_user/voxeldemo/main.c
+++ b/_user/voxeldemo/main.c
@@ -13,12 +13,17 @@
  * %LICENSE%
  */
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
+#include <signal.h>
 #include <math.h>
 #include <graph.h>
+
+
+volatile unsigned int flagQuit;
 
 
 typedef struct _camera_t {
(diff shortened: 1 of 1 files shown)


_user/voxeldemo/main.c
```diff
@@ -81,14 +84,19 @@ int voxel_init(voxel_t *v)
...
+			break;
+
+		v->width = g->width;
+		v->height = g->height;
 
-		return 0;
+		return EOK;
 	} while (0);
 
 	voxel_free(v);
```
kemonats, line 102: What value will have the `v->mapcolor variable`
if  `(v->mapHeight = malloc(1024 * 1024)) == NULL` will be true?
You should set variables before allocating memory.
```
v->mapColor = NULL;
v->bufback = NULL;
v->pixels = NULL;
v->palette = NULL;
```

gerard5: It is expected to set `voxel_t` structure/context properly before `voxel_init(..)` is called, thus:
```
	voxel_t v = {
		.cam = { .h = 300, .horiz = 200, .dist = 300 },
	};
```
is enough, any other member of the `voxel_t` structure is/are set to `NULL`, `0` by default, so all `mapHeight`, `mapColor`, `backBuf`, `pixels` and `palette` are set to `NULL` by default, (this is property of C initializer), before call to `voxel_init(..)`.

So, to answer Your question `mapHeight` will be `NULL`, but `mapColor` will be assigned a valid pointer returned by successful `malloc`, later `voxel_free(v)` is called and function returns an error, program fail then, and exits.

pawelpisarczyk:
Looks good to me./

pawelpisarczyk (approved):
Looks good to me.

PR #137, opened 2021-07-08
_targets/zynq7000: create ramscript.plo

HBuczynski:
## Description

At this stage, there is used only plo-ram-armv7a9-zynq700.img. After the last changes,
this image needs ramscript.plo.

JIRA: PD-151

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (qemu-zynq7000).

## Checklist:

## Special treatment

### DIFF
--- a/_targets/build.project.armv7a9-zynq7000
+++ b/_targets/build.project.armv7a9-zynq7000
@@ -19,11 +19,13 @@ PREINIT_SCRIPT=("map ddr 0x100000 0x900000 rwx"
 
 b_mkscript_preinit() {
 	mkdir -p "$PREFIX_BUILD/plo/"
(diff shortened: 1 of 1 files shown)


nalajcie (approved):
would be better to write script to temporary file, diff against the previous one and overwrite only if needed - to avoid unnecessary `plo` re-linking in incremental build scenario.

Current stop-gap LGTM.

HBuczynski:
> would be better to write script to temporary file, diff against the previous one and overwrite only if needed - to avoid unnecessary `plo` re-linking in incremental build scenario.
> 
> Current stop-gap LGTM.

Thanks, I will take your comment into account when next changes will be made in plo building process.

PR #174, opened 2021-08-20
update submodules

nalajcie:
* phoenix-rtos-ports 425fc1d...7ee7b2f (3):
  > busybox: make watchdog compilable
  > wpa_supplicant: disable some unused features
  > sscep: add mime patch

* phoenix-rtos-build 38a23eb...e9c5555 (3):
  > toolchain: make crtstuff relocatable on Cortex-M targets
  > toolchain: disable libitm support
  > toolchain: remove obsolete patches

**phoenix-rtos-kernel** submodule is not being updated due to #176


nalajcie:
@jsarzy apart from `setjmp` test failing, the test runner throws exception while trying to parse the output, can You check it at a later time (the runner part, the failing tests is probably due to https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/227, I hope @agkaminski would be able to check it).

jsarzy:
`test-setjmp` (and I bet `test-libc`) produces BUS_FAULT exception and causes device reboot. On restart some weird bytes are read, that is why runner can't produce a helpful output. I think we can add catching this exception on harness call and tries to print what we have read. I will create an issue.

The output of failed test - it seems like execution of Unity causes the exception:
image

nalajcie:
as both @jsarzy and @agkaminski  are AFK this week, I'm dropping the `phoenix-rtos-kernel` submodule update.

@gerard5 If You would be working on `rt1064` target this week, can You try to narrow it down to one of the commits?

anglov (approved):
LGTM

gerard5:
> as both @jsarzy and @agkaminski are AFK this week, I'm dropping the `phoenix-rtos-kernel` submodule update.
> 
> @gerard5 If You would be working on `rt1064` target this week, can You try to narrow it down to one of the commits?

I will try to (on `imxrt1064`), and as I'm working more on `imxrt1176`  I may also check `setjmp` on it (and if it worked before).

PR #367, opened 2022-03-25
scripts: add ia32-generic script with net setup for testing purposes, specifying pentium3 cpu in all scripts

damianloew:
## Description

1) scripts: add ia32-generic script with net setup for testing purposes
2) scripts: ia32-generic: set cpu to supported pentium3

## Motivation and Context

1) JIRA: CI-116
2) JIRA: PD-251

1) The main purpose is to enable proper lwip launching in CI process by updating qemu running command.

Now the command for running qemu (ia32-generic target) is now placed strictly in test runner:
Screenshot from 2022-03-25 12-30-04

It may be better to create the proper script in `scripts` directory that could be used by a test runner. If there is a need for change in qemu running it will be done only in the `scripts` directory (it's easier to remember about changing similar script in the same directory than changing some file in tests repository).

2) When some ca-certificates are used by openssl, sse2 instructions may be executed.
Openssl firstly checks wheather an extension exists, but when it does and isn't supported in kernel, illegal instructions are executed.
sse2 can't also be disabled (on processors with it, e. g. pentium4), even if it's not supported by the os.
That's why the cpu is set to supported one in ia32-generic scripts - pentium3.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic qemu).

CI process passes with this change and https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/104 where the script is used in test runner

## Checklist:

## Special treatment

### DIFF
--- a/scripts/ia32-generic-test.sh
+++ b/scripts/ia32-generic-test.sh
@@ -7,6 +7,7 @@
 #
 
 exec qemu-system-i386 \
+	-cpu pentium3 \
 	-hda "$(dirname "${BASH_SOURCE[0]}")/../_boot/phoenix-ia32-generic.disk" \
 	-nographic \
 	-monitor none \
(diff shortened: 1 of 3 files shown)


niewim19 (approved):
LGTM

PR #413, opened 2022-06-06
Introduce new projects configuration and build structure

Maxez:
## Description

1. Every project name has a format `TARGET_FAMILY-TARGET_SUBFAMILY-TARGET_PROJECT` and defines new `TARGET`. The project has a separate folder with its full name (`TARGET`) inside **_projects/** (new `PROJECT_PATH` variable points to that folder). Inside that folder there are required `build.project` and `board_config.h` files.
2. `build.project*` files inside **_targets/** no longer define valid `TARGET`s - they're templates used by `build.project` files inside `PROJECT_PATH`.
3. `PROJECT_PATH` is added to `plo`, `kernel` and `devices` include paths to make `board_config.h` visible with:
```c
#include <board_config.h>
```
6. `TARGET`s `riscv64-virt` and `riscv64-spike` are replaced by `riscv64-generic-qemu` and `riscv64-generic-spike`.
7. New `armv7a9-zynq7000-zturn` `TARGET` is added.
5. List of available default `TARGET`s after changes:
`armv7a7-imx6ull-evk`
`armv7a9-zynq7000-qemu`
`armv7a9-zynq7000-zedboard`
`armv7a9-zynq7000-zturn`
`armv7m4-stm32l4x6-nucleo`
`armv7m7-imxrt105x-evk`
`armv7m7-imxrt106x-evk`
`armv7m7-imxrt117x-evk`
`host-generic-pc`
`ia32-generic-pc`
`ia32-generic-qemu`
`riscv64-generic-qemu`
`riscv64-generic-spike`

## Motivation and Context

[JIRA: PP-17](https://jira.phoenix-rtos.com/browse/PP-17)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
Existing projects will no longer build properly. They have to be adapted to the new structure.

## How Has This Been Tested?

## Checklist:

## Special treatment


_projects/host-generic-pc/build.project
```diff
@@ -0,0 +1,15 @@
...
+# Copyright 2022 Phoenix Systems
+# Author: Lukasz Kosinski
+#
+
+[ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
+
+. "_targets/build.project.host-generic"
+
+b_image_project () {
+	b_log "The images have been built for the host-generic-pc platform"
```
agkaminski, line 14: I think information about platform name can be fetched from variables, this will save some editing when making these files

anglov:
Some notes from me maybe to be addressed in the future. Do not treat it as a review:
- In general current approach looks good in terms of what defines what. Board configuration is always defined by a project, build recipe also. 
- I don't like that you named third part of target as TARGET_PROJECT. It's rather something like TARGET_PLATFORM, as you want to place qemu, evk etc. here. I think it's not in the code or docs anywhere so just pointing it as it placed in PR description.
- phoenix-rtos-project contains only one project (one binary for one target). It's reasonable but it'll be quite common to generate different firmware for the same device. Current phoenix-rtos-project approach disallows this.
- I don't like that target is main for project build – i think that in build.project should be b_bulild_project which is defined for each project. And those functions should call functions from _target directory. Then it allows projects to control the build things.

I like presented approach anyways and I think it makes step in good direction.

agkaminski:
> Some notes from me maybe to be addressed in the future. Do not treat it as a review:
> 
>     * In general current approach looks good in terms of what defines what. Board configuration is always defined by a project, build recipe also.
> 
>     * I don't like that you named third part of target as TARGET_PROJECT. It's rather something like TARGET_PLATFORM, as you want to place qemu, evk etc. here. I think it's not in the code or docs anywhere so just pointing it as it placed in PR description.
>       EDIT: correct self: TARGET_PROJECT is already used.
> 

evk, qemu etc are project names - phoenix-rtos-project contains example projects, these are names of them

>     * phoenix-rtos-project contains only one project (one binary for one target). It's reasonable but it'll be quite common to generate different firmware for the same device. Current phoenix-rtos-project approach disallows this.
> 

Yes, this is planned in the near future to allow good separation between artefacts (dir per project in _boot)

>     * I don't like that target is main for project build – i think that in build.project should be b_bulild_project which is defined for each project. And those functions should call functions from _target directory. Then it allows projects to control the build things.
> 
> 
> I like presented approach anyways and I think it makes step in good direction.

nalajcie:
> Some notes from me maybe to be addressed in the future. Do not treat it as a review:
> 
> * In general current approach looks good in terms of what defines what. Board configuration is always defined by a project, build recipe also.
> * I don't like that you named third part of target as TARGET_PROJECT. It's rather something like TARGET_PLATFORM, as you want to place qemu, evk etc. here. I think it's not in the code or docs anywhere so just pointing it as it placed in PR description.

The usual convension in other OSes is to use name `board` to specify particular HW instance (see for example linux / zephyr  / buildroot) or `device` (android / yocto). IMHO `project` is a misleading name - but I hope it's an evolution in the right direction. (see - the include is already named `board_config.h`, not `project_config.h`)

>   EDIT: correct self: TARGET_PROJECT is already used.
> * phoenix-rtos-project contains only one project (one binary for one target). It's reasonable but it'll be quite common to generate different firmware for the same device. Current phoenix-rtos-project approach disallows this.

My point exactly. We should be able to have multiple projects per board (not necessarily in  `phoenix-rtos-project` but it should be possible in our buildsystem).

> * I don't like that target is main for project build – i think that in build.project should be b_bulild_project which is defined for each project. And those functions should call functions from _target directory. Then it allows projects to control the build things.
> 
> I like presented approach anyways and I think it makes step in good direction.

agkaminski:
@nalajcie @anglov You assume that board config will be identical for all projects for a particular board - imho it will be not. E.g console required on a different UART, some expantion slot populated with different HW, different NV memory used (EVKs have selectable NV memories connected to the same pins) etc. It sounds good, but would prove non practical - we would need to remove some constrains from board_config.h and move them to project_config.h regularly or populate project with many virtual board configs for all needed options.

_projects/host-generic-pc/build.project
Maxez: Done, used TARGET variable.

agkaminski (approved):

PR #449, opened 2022-07-06
enable building ports within the CI process for armv7a9-zynq7000-qemu

damianloew:
## Description

- enable building ports within the CI process for armv7a9-zynq7000-qemu

## Motivation and Context

 - `armv7a9-zynq7000-qemu` target now runs the busybox ash on startup. To make running tests in CI possible busybox port has to be built.

 - JIRA: PD-291

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/.github/workflows/ci-submodule.yml
+++ b/.github/workflows/ci-submodule.yml
@@ -29,6 +29,8 @@ jobs:
         include:
           - target: 'ia32-generic-qemu'
             syspage: 'psh pc-ata uart16550'
+          - target: 'armv7a9-zynq7000-qemu'
+            additional_params: 'ports'
     steps:
       # step 1: checkout phoenix-rtos-project repository code inside the workspace directory of the runner
       - name: Checkout phoenix-rtos-project
@@ -55,7 +57,7 @@ jobs:
         with:
           target: ${{ matrix.target }}
           syspage: ${{ matrix.syspage }}
-          params: ${{ inputs.build_params }}
+          params: ${{ inputs.build_params }} ${{ matrix.additional_params }}
 
       # step 4: tar rootfs
       - name: Tar rootfs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,8 @@ jobs:
         include:
           - target: 'ia32-generic-qemu'
             syspage: 'psh pc-ata uart16550'
+          - target: 'armv7a9-zynq7000-qemu'
+            additional_params: 'ports'
     steps:
       # step 1: checkout repository code inside the workspace directory of the runner
       - name: Checkout the repository
@@ -42,7 +44,7 @@ jobs:
         with:
           target: ${{ matrix.target }}
           syspage: ${{ matrix.syspage }}
-          params: 'core fs test project image'
+          params: core fs test project image ${{ matrix.additional_params }}
 
       # step 3: tar rootfs
       - name: Tar rootfs


jsarzy (commented):
imho, I think it's OK to do it this way.

@nalajcie what do you think?

.github/workflows/ci-submodule.yml
```diff
@@ -25,10 +25,13 @@ jobs:
       build_result: ${{ steps.build.outcome }}
     strategy:
       matrix:
-        target: ['armv7a7-imx6ull-evk', 'armv7a9-zynq7000-qemu', 'armv7a9-zynq7000-zedboard', 'armv7a9-zynq7000-zturn', 'armv7m4-stm32l4x6-nucleo', 'armv7m7-imxrt106x-evk', 'armv7m7-imxrt117x-evk', 'host-generic-pc', 'ia32-generic-pc', 'ia32-generic-qemu', 'riscv64-generic-qemu', 'riscv64-generic-spike']
+        target: ['armv7a7-imx6ull-evk', 'armv7a9-zynq7000-zedboard', 'armv7a9-zynq7000-zturn', 'armv7m4-stm32l4x6-nucleo', 'armv7m7-imxrt106x-evk', 'armv7m7-imxrt117x-evk', 'host-generic-pc', 'ia32-generic-pc', 'ia32-generic-qemu', 'riscv64-generic-qemu', 'riscv64-generic-spike']
+        additional_params: ['']
```
jsarzy, line 29: Do we need this empty `additional_params`?

damianloew: Oh, that's right. It isn't necessary here, now it's updated.

jsarzy: Do we need to remove `armv7a9-zynq7000-qemu` from target list? `ia32-generic-qemu` is there and work wells with additional `syspage` param

damianloew: I think it could be specified in `target`. Maybe it's a better way, the change was applied in the last force push.

jsarzy: And last note, why we removed quotation marks? `''`

.github/workflows/ci.yml
```diff
@@ -42,7 +44,7 @@ jobs:
         with:
           target: ${{ matrix.target }}
           syspage: ${{ matrix.syspage }}
-          params: 'core fs test project image'
+          params: core fs test project image ${{ matrix.additional_params }}
```
jsarzy, line 47: why did we remove quotation marks?

damianloew: There is the following error if I did it this way
Screenshot from 2022-07-22 13-31-34

damianloew: I've applied the previous approach and squashed the commits.

jsarzy: Yeap but I think it may not work from ci-subcalls. If I understand how arguments are passed, we call the build script there as follows: `./phoenix-rtos-build/build.sh 'core fs test project image' 'test'`, and it will build only test in that case. I think we should just define default params in matrix and overwrite it in the `include` directive (as I remember that was the previous approach?)

damianloew: > Yeap but I think it may not work from ci-subcalls. If I understand how arguments are passed, we call the build script there as follows: `./phoenix-rtos-build/build.sh 'core fs test project image' 'ports'`, and it will build only test in that case. I think we should just define default params in matrix and overwrite it in the `include` directive (as I remember that was the previous approach?)

It works, I've re-triggered workflows here to be sure (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/118/checks). The previous attempt requires changing `required jobs` for all repos and if there were some changes in params we would have to update them (if I understand correctly)

damianloew: Yeah, it seems to be ok:
Screenshot from 2022-07-22 14-17-56
Screenshot from 2022-07-22 14-18-16

jsarzy (approved):

PR #537, opened 2022-10-26
add mbedtls and azure sdk to ports config

damianloew:
## Description

 - add mbedtls and azure sdk to ports config

## Motivation and Context

 - JIRA: PD-251

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic-qemu, imxrt106x).

## Checklist:

## Special treatment

### DIFF
--- a/_projects/ia32-generic-qemu/build.project
+++ b/_projects/ia32-generic-qemu/build.project
@@ -9,6 +9,13 @@
 [ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
 
 
+#
+# Ports configuration - additional to the one from _targets
+#
+export PORTS_MBEDTLS=y
+export PORTS_AZURE_SDK=y
+
+
 b_image_project () {
 	b_log "The images have been built for the ${TARGET} platform"
 }
--- a/_targets/build.project.armv7m7-imxrt106x
+++ b/_targets/build.project.armv7m7-imxrt106x
@@ -29,6 +29,8 @@ export PORTS_OPENVPN=n
 export PORTS_JANSSON=n
 export PORTS_CURL=n
 export PORTS_MICROPYTHON=n
+export PORTS_MBEDTLS=n
+export PORTS_AZURE_SDK=n
 
 
 #


agkaminski (commented):
Other platforms don't require `export PORTS_MBEDTLS=n` and `export PORTS_AZURE_SDK=n`?

btw, imho these defines should be considered to be `n` by default (need explicit enable to be enabled), so `=n` config would not be needed. If it's needed than all projects need update with this disable when new `ports` feature becomes available, which is unfortunate

damianloew:
> Other platforms don't require `export PORTS_MBEDTLS=n` and `export PORTS_AZURE_SDK=n`?
> 
> btw, imho these defines should be considered to be `n` by default (need explicit enable to be enabled), so `=n` config would not be needed. If it's needed than all projects need update with this disable when new `ports` feature becomes available, which is unfortunate

I thought that `export PORTS_PORTNAME=n` means that we can enable it if we want to, so I've just added those ports to platforms, where they were tested. Please let me know whether we want to set all ports for all platforms to `n` by default.

Personally, I would prefer leaving some default configuration for the specific target architecture in_`targets` and mentioning there all ports possible to run. Additionally, if we wanted to have some different ports config for specific project/board than the one related to its target architecture this would be placed in `_projects`.

nalajcie:
> Other platforms don't require `export PORTS_MBEDTLS=n` and `export PORTS_AZURE_SDK=n`?
> 
> btw, imho these defines should be considered to be `n` by default (need explicit enable to be enabled), so `=n` config would not be needed. If it's needed than all projects need update with this disable when new `ports` feature becomes available, which is unfortunate

All of the ports are disabled by default only explicitly enabling them would make them build (see https://github.com/phoenix-rtos/phoenix-rtos-ports/blob/b46f4344ad527bf55b6ab3ddfbf81eba7278292c/build.sh#L38).

I think the intention was to highlight that this port is supported for `rt106x` and could be enabled by the user if desired. Still, a comment explaining that should be added ;).

agkaminski:
Yeah, it is as I would like it to be - build only if `=y`, so it's ok.

agkaminski (approved):

PR #662, opened 2023-03-15
imx6ull: boot from plo on nor flash

badochov:
JIRA: RTOS-398

## Description
boot from plo on nor flash.

* phoenix-rtos-doc 4eaac00...5d1f42a (3):
  > libc: add getopt, getpgid, getsid, memccpy, setenv, strncpy, truncate
  > libc: documentation correction
  > imx6ull: boot from nor (#139)

* phoenix-rtos-hostutils 853e232...a56b518 (1):
  > imx6ull: boot from qspi nor flash

* plo 315360b...42bc591 (4):
  > imx6ull: flash nor driver over qspi
  > flash-imx6ull: Add optional logic to perform big buffer read.
  > cmds: implement test-dev
  > imx6ull: boot from qspi nor flash

## Motivation and Context
imx6ull lacks support for starting from flash

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## How Has This Been Tested?

- [ ] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [x] Tested by hand on: imx6ull.

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing linter checks and tests passed.
- [x] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [x] This PR needs additional PRs to work
https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/378
https://github.com/phoenix-rtos/plo/pull/231
https://github.com/phoenix-rtos/plo/pull/236
https://github.com/phoenix-rtos/phoenix-rtos-hostutils/pull/42
https://github.com/phoenix-rtos/phoenix-rtos-doc/pull/139
- [ ] I will merge this PR by myself when appropriate.

PR #726, opened 2023-05-12
imx6ull-evk: add JFFS2 rootfs

badochov:
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imx6ull

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work .
https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/369
https://github.com/phoenix-rtos/phoenix-rtos-project/pull/662
https://github.com/phoenix-rtos/phoenix-rtos-project/pull/728

### DIFF
--- a/_projects/armv7a7-imx6ull-evk/build.project
+++ b/_projects/armv7a7-imx6ull-evk/build.project
@@ -2,8 +2,8 @@
 #
 # Shell script for building armv7a7-imx6ull-evk project
 #
-# Copyright 2022 Phoenix Systems
-# Author: Lukasz Kosinski
+# Copyright 2022-2023 Phoenix Systems
+# Author: Lukasz Kosinski, Hubert Badocha
 #
 
 [ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
@@ -26,7 +26,16 @@ BOOT_DEVICE="nor0"                # Default boot device
 OFFS_USER_SCRIPT=$((0x20000))     # Disk image user script offset (in bytes)
 
 KERNEL_OFFS=$((0x21000))          # Disk image kernel offset (in bytes)
+ERASE_SIZE=$((16 * 4096))         # Rootfs disk erase block size (in bytes)
+SECTOR_SIZE=$((0x100))            # Rootfs write sector size (in bytes)
+CLEANMARKER_SIZE=$((0x10))        # JFFS2 cleanmarker size (in bytes)
+IMAGE_PAD_FF=y
 
+#
+# Fs data
+#
+FS_OFFS=$((0x100000))
+FS_SZ=$((0x2000000 - FS_OFFS))
 
 # Pre-init script is launched before user script
 PREINIT_SCRIPT=(
@@ -37,7 +46,7 @@ PREINIT_SCRIPT=(
 )
 
 
-#TODO: if flash driver for imx6ull-evk implementation is introduced, psh should be replaced by imx6ul-flash and psd
+#TODO: if flash driver for imx6ull-evk implementation is introduced, psh should be replaced by imx6ull-flash and psd
 # This is temporary solution for running a system image in RAM
 UPLOAD_SCRIPT=(
 	"wait 2000"
@@ -52,13 +61,17 @@ UPLOAD_SCRIPT=(
 PRIMARY_SCRIPT=(
 	"wait 2000"
 	"kernel ${BOOT_DEVICE}"
-	"app ${BOOT_DEVICE} -x dummyfs ddr ddr"
+	"app ${BOOT_DEVICE} -x dummyfs;-N;devfs;-D ddr ddr"
 	"app ${BOOT_DEVICE} -x imx6ull-uart ddr ddr"
-	"app ${BOOT_DEVICE} -x psh ddr ddr"
+	"app ${BOOT_DEVICE} -x imx6ull-flashnor;-q;1;-r;0;${FS_OFFS};${FS_SZ};jffs2 ddr ddr"
+	"app ${BOOT_DEVICE} -x psh;-i;/etc/rc.psh ddr ddr"
 	"go!"
 )
 
 
 b_image_project () {
+	OFFSET="$FS_OFFS"
+	b_add2img "$(b_rootfs_path_get_target)" "$PHOENIX_DISK"
+
 	b_log "The images have been built for the ${TARGET} platform"
 }
--- a/_targets/build.common
+++ b/_targets/build.common
@@ -126,17 +126,54 @@ b_mkscript_preinit() {
 
 
 # Function gets the following arguments:
+# - $1 - path to image file
+# - $2 - pad size
+# When variable IMAGE_PAD_FF is set to "y" parts of the images get padded with 0xff.
+b_pad()
+{
+	if [ "$2" -ne 0 ]; then
+		if [ "$IMAGE_PAD_FF" = 'y' ]; then
+			dd if=/dev/zero bs="$2" count=1 2>/dev/null | tr "\000" "\377" > padding.img
+			dd if="$1" of=padding.img conv=notrunc 2>/dev/null
+			mv padding.img "$1"
+		else
+			truncate -s "$2" "$1"
+		fi
+	fi
+}
+
+
+# Function gets the following arguments:
 # - $1 - path to file which should be added to image
 # - $2 - path to image file
+# When variable IMAGE_PAD_FF is set to "y" parts of the images get padded with 0xff.
 b_add2img() {
 	# Function needs external variables to be defined in a project file:
 	: "${OFFSET:?variable unset}"    # starting file's position in the image
 	: "${SIZE_PAGE:?variable unset}"
 	: "${PAGE_MASK:?variable unset}"
 
-	printf "Copying %s (offs=%dB)\n" "$1" "$OFFSET"
-	dd if="$1" of="$2" bs=1 seek="$OFFSET" conv=notrunc 2>/dev/null
-	sz=$((($(wc -c < "$1") + SIZE_PAGE - 1) & PAGE_MASK))
+	if [ "$IMAGE_PAD_FF" = "y" ]; then
+		dd_pad_str=", pad=0xff"
+	else
+		dd_pad_str=""
+	fi
+
+	sz=$(wc -c < "$1")
+
+	printf "Copying %-100s (size=0x%06xB, offs=0x%08x%s) to %s\n" "$1" "$sz" "$OFFSET" "$dd_pad_str" "$(basename "$2")"
+
+	touch "$2"
+	if [ "$OFFSET" -ge "$(wc -c < "$2")" ]; then
+		# Fill the space between parts of the image with 0xff.
+		b_pad "$2" "$OFFSET"
+		cat "$1" >> "$2"
+	else
+		# Writing in the middle of the image. Have to use slower dd.
+		dd if="$1" of="$2" bs=1 seek="$OFFSET" conv=notrunc 2>/dev/null
+	fi
+
+	sz=$(((sz + SIZE_PAGE - 1) & PAGE_MASK))
 	OFFSET=$((OFFSET + sz))
 }
 
@@ -175,7 +212,7 @@ b_prod_image() {
 	done
 
 	# System image (without plo)
-	dd if="$PHOENIX_DISK" of="$PHOENIX_IMG" bs=1 skip="$OFFS_USER_SCRIPT" 2>/dev/null
+	tail -c +$((OFFS_USER_SCRIPT + 1)) "$PHOENIX_DISK" > "$PHOENIX_IMG"
 }
 
 
@@ -226,8 +263,8 @@ b_dev_image() {
 	OFFSET="$OFFS_USER_SCRIPT"
 	b_add2img "$PLO_SCRIPT_DIR/$NAME_USER_SCRIPT" "$PLO_DEV_IMG"
 
-	dd if=/dev/zero of="$PLO_DEV_IMG" bs=1 seek="$OFFSET" count=$((KERNEL_OFFS - OFFS_USER_SCRIPT)) 2>/dev/null
-
+	# TODO is this resize really needed?
+	b_pad "$KERNEL_OFFS" "$PLO_DEV_IMG"
 
 	# Dev user script
 	OFFSET=0
--- a/_targets/build.project.armv7a7-imx6ull
+++ b/_targets/build.project.armv7a7-imx6ull
@@ -2,8 +2,8 @@
 #
 # Shell script for building armv7a7-imx6ull project
 #
-# Copyright 2018, 2019, 2020, 2021 Phoenix Systems
-# Author: Kaja Swat, Aleksander Kaminski, Pawel Pisarczyk, Lukasz Kosinski, Maciej Purski
+# Copyright 2018, 2019, 2020, 2021, 2023 Phoenix Systems
+# Author: Kaja Swat, Aleksander Kaminski, Pawel Pisarczyk, Lukasz Kosinski, Maciej Purski, Hubert Badocha
 #
 [ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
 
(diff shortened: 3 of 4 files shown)


agkaminski (commented):
LGTM, adding @nalajcie to be sure

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
 	# FIXME: does work only on macOS, for now hackish solution based on extending PATH
 	mtd-utils/build.sh
 	PATH="$(pwd)/_build/host-generic-pc/prog/:$PATH"
-	ROOTFS="$PREFIX_BOOT/rootfs.jffs2"
+	export ROOTFS="$PREFIX_BOOT/rootfs.jffs2"
+
+	if [[ $CLEANMARKER_SIZE = 0 ]]; then
+		CLEANMARKER_ARG=-n
+	else
+		CLEANMARKER_ARG="-c $CLEANMARKER_SIZE"
```
nalajcie, line 165: make it a bash array for correct parameter quoting below

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
+		b_kernel_image "$PREFIX_BOOT/phoenix-kernel.img"
+		b_syspage_gen "$PREFIX_BOOT/phoenix-kernel.img" "$PLO_SCRIPT_DIR/$NAME_USER_SCRIPT"
+	fi
 
 	b_log "Creating rootfs"
 	# FIXME: does work only on macOS, for now hackish solution based on extending PATH
 	mtd-utils/build.sh
 	PATH="$(pwd)/_build/host-generic-pc/prog/:$PATH"
-	ROOTFS="$PREFIX_BOOT/rootfs.jffs2"
+	export ROOTFS="$PREFIX_BOOT/rootfs.jffs2"
```
nalajcie, line 160: why exporting? not used by any of the subprocesses? If you want to reuse it in `b_image_project` then maybe define in global scope instead of depending on order of function calling?

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
+
+	if [[ $CLEANMARKER_SIZE = 0 ]]; then
+		CLEANMARKER_ARG=-n
+	else
+		CLEANMARKER_ARG="-c $CLEANMARKER_SIZE"
+	fi
+
+	mkfs.jffs2 -U $CLEANMARKER_ARG -l -m none -e $ERASE_SIZE -s $SECTOR_SIZE -r "$PREFIX_ROOTFS"/ -o "$ROOTFS" -p
```
nalajcie, line 168: not sure if `-p` was not specifically omitted previously (writing `0xFF` bytes to NAND is not a no-op - it changes ECC bytes) - please verify if it's correctly skipped by `nandwrite`

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
 
-	mkfs.jffs2 -U -m none -e $((64*4096)) -s 4096 -n -r "$PREFIX_ROOTFS"/ -o "$ROOTFS"
-	if sumtool -e $((64*4096)) -i "$ROOTFS" -o "$ROOTFS.tmp" 2> /dev/null; then
+	if sumtool $CLEANMARKER_ARG -e $ERASE_SIZE -l -i "$ROOTFS" -o "$ROOTFS.tmp" -p 2> /dev/null; then
```
nalajcie, line 170: verify if `-p` won't break NAND writing

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
-	b_log "RAM Image"
-	b_kernel_image "$PREFIX_BOOT/phoenix.img"
-	b_syspage_gen "$PREFIX_BOOT/phoenix.img" "$PLO_SCRIPT_DIR/$NAME_USER_SCRIPT"
 
-	b_log "FLASH Image"
-	b_mkscript_user "${PRIMARY_SCRIPT[@]}"
-	b_kernel_image "$PREFIX_BOOT/phoenix-kernel.img"
-	b_syspage_gen "$PREFIX_BOOT/phoenix-kernel.img" "$PLO_SCRIPT_DIR/$NAME_USER_SCRIPT"
+	if [ "$KERNEL_PLO_BOOT" = "y" ]; then
+		b_log "Image will boot from PLO."
```
nalajcie, line 132: IMHO the plo should be part of the disk image (I'm assuming currently it's not?) Probably kernel offset in the disk image would need to be changed, not sure if we want to introduce `ptable` at this stage or not

_targets/build.project.armv7a7-imx6ull
```diff
@@ -113,26 +127,47 @@ b_image_target() {
...
+		# Parsed by syspagen for map and console commands.
+		b_mkscript_preinit
+
+		# Build production image
+		b_log "RAM Image"
+		b_kernel_image "$PREFIX_BOOT/phoenix.img"
+		b_syspage_gen "$PREFIX_BOOT/phoenix.img" "$PLO_SCRIPT_DIR/$NAME_USER_SCRIPT"
+
+		b_log "FLASH Image"
+		# This is a dirty workaround.
```
nalajcie, line 147: IMHO not a valid comment as it's used in production :P. This is a workaround for targets not using PLO as a second stage bootloader on 6ull - and it's used to produce kernel image with required syspage apps embedded (not only uploading, this image resides on NAND and is treated as a single blob by BootROM)

_targets/build.project.armv7a7-imx6ull
```diff
@@ -34,46 +34,51 @@ export PORTS_CURL=y
...
 #
 # Project specific build
 #
-
 BOOT_DEVICE="flash0"              # Default boot device
 NAME_USER_SCRIPT="user.plo"       # User script name
 MAGIC_USER_SCRIPT=$((0xdabaabad)) # User script magic value
 OFFS_USER_SCRIPT=$((0x0))         # Disk image user script offset (in bytes)
 OFFS_ROOTFS=$((0x1000))           # Disk image rootfs offset (in 512 byte sectors)
+ERASE_SIZE=$((0x40000))           # Rootfs disk erase block size (in bytes)
```
nalajcie, line 49: IMHO `4096` and `64 * 4096`  are more readable instead of hex numbers

nalajcie (changes requested):
marking as at **request changes** due to smuggling `-p` parameter - not sure if it wasn't omitted on purpose - please confirm the functionality on ECC NAND (we don't want to actually write `0xff` sectors in the last erase block as this will write non-zero ECC bytes and make the erase block unusable.

Also:
 - some nitpicks
 - please do submodule update in separate commit (with common commit message highlighting the changes)

Haven't tested on HW.

_targets/build.project.armv7a7-imx6ull
badochov: why would like to have PLO on targets not using it?

_targets/build.project.armv7a7-imx6ull
badochov: decided to remove `-p` as the method for flashing on NOR erases the JFFS2 sectors on its own

_targets/build.project.armv7a7-imx6ull
badochov: I agree that writing full disk images is wasteful. However I am not sure how big of an issue it is as the gaps between parts should be as small as possible both in full images and in splitted smaller images, so  amount of zeroed block shouldn't be higher that few. @agkaminski what's your take on that matter?

_targets/build.project.armv7a7-imx6ull
badochov: $PREFIX_BOOT is not defined in the top level as this file is sources before it is set. Maybe we should change the order in the buiild script, however for now I don't see a better way of passing ROOTFS

nalajcie:
Please note that recent changes introduced multiple new compilation warnings - please solve them as soon as possible and take care in the next PRs not to introduce more unnecessary ones.

Compiling imx6ull-evk:
image

makefile errors:
image

 - installing header files twice in different components (`LOCAL_HEDERS` is for installing particular component external interface, heder dependencies will be dealt automatically by gcc dependency generation)
 - please use `include` subdir to automatically install headers subtree if needed
 - header files are platform specific - you can't use generic names (like `nor.h`) - either make it a generic interface or change the names to platform-specific (eg. `imx6ull-nor.h`)

_targets/build.project.armv7a7-imx6ull
nalajcie: After some more thinking:
 - the disk image can't realistically be used for NAND use case (we can't write `0xff` as a no-op, we need to check/skip badblocks, etc.) so maybe let's not focus on that (should we add some comment?)
 - for NOR `disk` image for sure is useful, but all free space should be marked as `0xff` - the paddings might be considerable (if ptable would be added to the disk image - it usually resides at the end of the flash)

Sample code for `0xff` padding generation (can be done in other ways):
```bash
pad_ff()
{
	dd if=/dev/zero ibs=1 count="$2" 2>/dev/null | tr "\000" "\377" > padding.img
	dd if="$1" of=padding.img conv=notrunc 2>/dev/null
	mv padding.img "$1"
}
```

badochov:
> Please note that recent changes introduced multiple new compilation warnings - please solve them as soon as possible and take care in the next PRs not to introduce more unnecessary ones.
> 
> Compiling imx6ull-evk: image
> 
> makefile errors: image
> 
> * installing header files twice in different components (`LOCAL_HEDERS` is for installing particular component external interface, heder dependencies will be dealt automatically by gcc dependency generation)
> * please use `include` subdir to automatically install headers subtree if needed
> * header files are platform specific - you can't use generic names (like `nor.h`) - either make it a generic interface or change the names to platform-specific (eg. `imx6ull-nor.h`)

Please update devices to master, those warnings shouldn't occur on it

nalajcie:
> > Please note that recent changes introduced multiple new compilation warnings - please solve them as soon as possible and take care in the next PRs not to introduce more unnecessary ones.
> > Compiling imx6ull-evk: image
> > makefile errors: image
> > 
> > * installing header files twice in different components (`LOCAL_HEDERS` is for installing particular component external interface, heder dependencies will be dealt automatically by gcc dependency generation)
> > * please use `include` subdir to automatically install headers subtree if needed
> > * header files are platform specific - you can't use generic names (like `nor.h`) - either make it a generic interface or change the names to platform-specific (eg. `imx6ull-nor.h`)
> 
> Please update devices to master, those warnings shouldn't occur on it

Ack, the warnings are gone, still the other comments are valid:
 - consider if all othe installed headers are indeed external component interface - if not, use them directly in implementation and don't install it
 - all installed platform-specific headers need to have platform stem to avoid name collisions in the future (eg. You can place them in `include/imx6ull-flashnor/file.h` so they will be installed as `imx6ull-flashnor/fille.h` or maybe you just need to install one external API file (`imx6ull-flashnor.h`)

_projects/armv7a7-imx6ull-evk/build.project
```diff
@@ -26,7 +26,16 @@ BOOT_DEVICE="nor0"                # Default boot device
 OFFS_USER_SCRIPT=$((0x20000))     # Disk image user script offset (in bytes)
 
 KERNEL_OFFS=$((0x21000))          # Disk image kernel offset (in bytes)
+ERASE_SIZE=$((16 * 4096))         # Rootfs disk erase block size (in bytes)
+SECTOR_SIZE=$((0x100))            # Rootfs write sector size (in bytes)
+CLEANMARKER_SIZE=$((0x10))        # JFFS2 cleanmarker size (in bytes)
+FLASH_IMAGE=y
```
nalajcie, line 32: hmm, maybe name it better (`IMAGE_PAD_FF`?)  or at least write a comment what it means (`# disk image: pad with 0xFF (necessary for flash memory)`?)

Edit: maybe more detailed comment about the meaning should be in `b_add2img`

_targets/build.common
```diff
@@ -8,6 +8,7 @@
 [ "${BASH_SOURCE[0]}" -ef "$0" ] && echo "You should source this script, not execute it!" && exit 1
 
 KERNEL_ELF="phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf"
+FLASH_IMAGE=n
```
nalajcie, line 11: uh, failing in `b_add2img` if `FLASH_IMAGE` not set but setting it globally for every target? Please decide on one approach.

_targets/build.common
```diff
@@ -133,9 +147,17 @@ b_add2img() {
 	: "${OFFSET:?variable unset}"    # starting file's position in the image
 	: "${SIZE_PAGE:?variable unset}"
 	: "${PAGE_MASK:?variable unset}"
+	: "${FLASH_IMAGE:?variable unset}"
```
nalajcie, line 150: please describe the meaning and either make it an optional with default val or set per-project

_targets/build.common
```diff
@@ -133,9 +147,17 @@ b_add2img() {
...
 
 	printf "Copying %s (offs=%dB)\n" "$1" "$OFFSET"
-	dd if="$1" of="$2" bs=1 seek="$OFFSET" conv=notrunc 2>/dev/null
+	if [ "$FLASH_IMAGE" = 'y' ]; then
+		b_pad_ff "$2" "$OFFSET"
+		touch "$2"
```
nalajcie, line 155: - why touch is needed?
- the file copying could be also done by the same `dd` (outside of if-else) - that would highlight that the `FLASH_IMAGE` param changes only padding value
- I would stick to one `printf` for consistency - regardless of the input params. The `printf` could obviously be extended, I have eg. `printf "Copying %-100s (offs=0x%08x) to %s\n" "$1" "$OFFSET" "$(basename "$2")"` in stash. Probably info about added padding size could also be beneficial)

_targets/build.project.armv7a7-imx6ull
nalajcie: hmm, maybe add some comment as it seems redundant. Also `ROOTFS` seems like a too generic name, maybe `ROOTFS_IMAGE` to prevent some `env` accidents?

nalajcie (commented):
the code looks much better now, leaving some nitpicks

_targets/build.common
nalajcie: 1. you can safely redirect to nonexistent file, shell will open(O_CREAT) it for you (eg. `cat /etc/passwd >> /tmp/nonexistent` works fine
2. I think `bs=1` is not needed here (dd will make the last block smaller without an issue) - using `cat` + redirect is AFAIK similar to using `bs=4096` (my personal preference would be to use `dd` - that way it's possible to add something in the middle /at the beginning of the image - now you can only append at the end)

nalajcie (commented):
- one typo found
- please note that the recent changes were (mistakenly) added to "add scripts to start kernel" commit
- not sure if needed, but as `b_add2img` invocations need to be monotonic maybe we should test it (and fail in case of overlapping offsets (`if [ "$OFFSET" -lt "$(wc -c "$1")" ]; then FAIL; fi`)

Apart from that, lgtm.

badochov:
> * one typo found
> * please note that the recent changes were (mistakenly) added to "add scripts to start kernel" commit
> * not sure if needed, but as `b_add2img` invocations need to be monotonic maybe we should test it (and fail in case of overlapping offsets (`if [ "$OFFSET" -lt "$(wc -c "$1")" ]; then FAIL; fi`)
> 
> Apart from that, lgtm.

when it comes to monitonicity of add2img I had to adjsut the code to non monotonic invocation due to this line in ia32 target
https://github.com/phoenix-rtos/phoenix-rtos-project/blob/0494ae4300dd195e84696732434e949cad87e6f8/_targets/build.project.ia32-generic#L140
possibly ia32 target can be adjusted, however,  I think it should be done in a separate PR if possible.

agkaminski (dismissed):
LGTM

_targets/build.project.armv7a7-imx6ull
```diff
@@ -153,10 +163,18 @@ b_image_target() {
 	# FIXME: does work only on macOS, for now hackish solution based on extending PATH
 	mtd-utils/build.sh
 	PATH="$(pwd)/_build/host-generic-pc/prog/:$PATH"
-	ROOTFS="$PREFIX_BOOT/rootfs.jffs2"
 
-	mkfs.jffs2 -U -m none -e $((64*4096)) -s 4096 -n -r "$PREFIX_ROOTFS"/ -o "$ROOTFS"
-	if sumtool -e $((64*4096)) -i "$ROOTFS" -o "$ROOTFS.tmp" 2> /dev/null; then
+	ROOTFS=$(b_rootfs_path_get_target)
+
+	if [[ $CLEANMARKER_SIZE = 0 ]]; then
```
nalajcie, line 169: nit: no need for compound commands, simple builtin is enough

```suggestion
	if [ "$CLEANMARKER_SIZE" -eq 0 ]; then
```

nalajcie (approved):

nalajcie (approved):

PR #760, opened 2023-06-21
project: add exmple of user app in cpp

DawidSzpejna:
## Description

project: add exmple of user app in cpp

## Motivation and Context

Presenting C++ in our system

JIRA: RTOS-405

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/_user/hellocpp/main.cpp
+++ b/_user/hellocpp/main.cpp
@@ -1,0 +1,24 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Hello World++
+ *
+ * Example of user application
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Dawid Szpejna
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <iostream>
+
+
+int main(void)
+{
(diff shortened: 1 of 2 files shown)


_user/hellocpp/main.cpp
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * %LICENSE%
+ */
+
+#include <iostream>
+
+
+int main(void)
+{
+	std::cout << "Hello World Cpp!" << std::endl;
```
damianloew, line 21: In my opinion it would be better, (eventually if you want to use cpp phrase there I'd rather print sth like `"Hello CppWorld!"`)

```suggestion
	std::cout << "Hello World!" << std::endl;
```

damianloew (commented):
Thanks for adding this demo, we could merge it in this version, but let's have it consistent.

_user/hellocpp/main.cpp
agkaminski: How about `Hello World++!`? :smile:

DawidSzpejna: @damianloew I corrected the string

agkaminski (approved):

PR #828, opened 2023-09-11
mtd-utils: update build.sh to a new convention

damianloew:
## Motivation and Context

fix building zynq/ull targets on macOS after changes in phoenix-rtos-build/build.sh

JIRA: CI-340

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [ ] This PR needs additional PRs to work ().

### DIFF
--- a/mtd-utils/build.sh
+++ b/mtd-utils/build.sh
@@ -6,12 +6,12 @@ if [ "$(uname)" = "Darwin" ]; then
 	b_log "Building mtd-utils"
 
 	PREFIX_MTD_UTILS="${PREFIX_PROJECT}/mtd-utils"
-	PREFIX_MTD_UTILS_BUILD="${PREFIX_PROJECT}/_build/host-generic-pc/mtd-utils"
+	PREFIX_MTD_UTILS_BUILD="${PREFIX_BUILD_HOST}/mtd-utils"
 
 	mkdir -p "$PREFIX_MTD_UTILS_BUILD"
 
 	pushd "$PREFIX_MTD_UTILS_BUILD"
-	if [ -n "$CLEAN" ]; then
+	if [ ! -f "${PREFIX_BUILD_HOST}/prog.stripped/mkfs.jffs2" ]; then
 		tar xzvf "$PREFIX_MTD_UTILS/mtd-utils-patched.tar.gz"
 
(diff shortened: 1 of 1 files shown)


mtd-utils/build.sh
```diff
@@ -11,7 +11,7 @@ if [ "$(uname)" = "Darwin" ]; then
 	mkdir -p "$PREFIX_MTD_UTILS_BUILD"
 
 	pushd "$PREFIX_MTD_UTILS_BUILD"
-	if [ -n "$CLEAN" ]; then
+	if [ "$B_CLEAN" = "y" ]; then
```
nalajcie, line 14: this should not depend on `B_CLEAN` variable but on existence of target files

nalajcie (changes requested):
Thanks for noticing this.

As this is a host tool I would gladly move this step to host preparation in docs and remove mtd-utils from `phoenix-rtos-project` altogether.

From what I see - getting `mkfs.jffs2` on Darwin is problematic so we can leave it as-is for now, but IMHO we should at least move it to hostsutils repo in the near future.

mtd-utils/build.sh
damianloew: Seems to be a better solution - updated. I'd merge it as it is for now.

mtd-utils/build.sh
```diff
@@ -7,11 +7,12 @@ if [ "$(uname)" = "Darwin" ]; then
 
 	PREFIX_MTD_UTILS="${PREFIX_PROJECT}/mtd-utils"
 	PREFIX_MTD_UTILS_BUILD="${PREFIX_PROJECT}/_build/host-generic-pc/mtd-utils"
```
nalajcie, line 9: maybe use `PREFIX_BUILD_HOST` (here and everywhere below also)?

damianloew: Good idea :+1:

nalajcie: `PREFIX_BUILD_HOST` is being exported from `build.sh`, don't define it by yourself. 

Also - missed the line above.

damianloew: Thanks, done.

nalajcie (approved):
Ok as a stop-gap for now.

PR #832, opened 2023-09-13
docker-build.sh: pass LONG_TEST variable

badochov:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -13,15 +13,15 @@ TMPFS_OVERLAY=()
 
 if [ "$#" -eq 1 ] && [ "$1" = "bash" ]; then
     # run interactive shell - using ROOT user
-    exec docker run -it --rm -v "$(pwd):/src" -w /src -e TARGET -e SYSPAGE -e CONSOLE --entrypoint bash $DOCKER_IMG_NAME
+    exec docker run -it --rm -v "$(pwd):/src" -w /src -e TARGET -e SYSPAGE -e CONSOLE -e LONG_TEST --entrypoint bash "$DOCKER_IMG_NAME"
 else
     # FIXME: run build - use our own UID/GID to create files with correct owner
     #exec docker run -it --user "$DOCKER_USER" --rm -v "$(pwd):/src:delegated" -w /src "${TMPFS_OVERLAY[@]}" -e TARGET -e SYSPAGE -e CONSOLE $DOCKER_IMG_NAME "$@"
 
     # FOR NOW: run build as root to be able to overwrite files installed in toolchain
-    docker run -it --rm -v "$(pwd):/src:delegated" -w /src "${TMPFS_OVERLAY[@]}" -e TARGET -e SYSPAGE -e CONSOLE $DOCKER_IMG_NAME "$@"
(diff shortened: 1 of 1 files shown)


badochov:
Are there any other potentially missing env variables?

nalajcie (approved):

damianloew (commented):
I think that `SYSPAGE` could be removed, but not necessarily within this PR

nalajcie:
> I think that `SYSPAGE` could be removed, but not necessarily within this PR

`SYSPAGE` is needed if you want to modify syspage, eg. build ia32-generc-qemu with serial UART instead of `vga` (you need to pass `CONSOLE=serial SYSPAGE='psh pc-ata uart16550' TARGET=ia32-generic-qemu`).

Please don't remove anything.

damianloew:
> > I think that `SYSPAGE` could be removed, but not necessarily within this PR
> 
> `SYSPAGE` is needed if you want to modify syspage, eg. build ia32-generc-qemu with serial UART instead of `vga` (you need to pass `CONSOLE=serial SYSPAGE='psh pc-ata uart16550' TARGET=ia32-generic-qemu`).
> 
> Please don't remove anything.



> uart16550

It seems that it's deprecated and no longer needed thanks to: https://github.com/phoenix-rtos/phoenix-rtos-project/blob/3cdb486e4276b9ac1fa8ff88e840c38d6939b665/_targets/build.project.ia32-generic#L65

Please correct me if I am wrong (setting CONSOLE=serial is sufficient when building on Ubuntu host)

damianloew:
> > > > I think that `SYSPAGE` could be removed, but not necessarily within this PR
> > > 
> > > 
> > > `SYSPAGE` is needed if you want to modify syspage, eg. build ia32-generc-qemu with serial UART instead of `vga` (you need to pass `CONSOLE=serial SYSPAGE='psh pc-ata uart16550' TARGET=ia32-generic-qemu`).
> > > Please don't remove anything.
> > 
> > 
> > > uart16550
> > 
> > 
> > It seems that it's deprecated and no longer needed thanks to:
> > https://github.com/phoenix-rtos/phoenix-rtos-project/blob/3cdb486e4276b9ac1fa8ff88e840c38d6939b665/_targets/build.project.ia32-generic#L65
> > 
> > Please correct me if I am wrong (setting CONSOLE=serial is sufficient when building on Ubuntu host)
> 
> I stand corrected, not needed in this particular use case. Still, `SYSPAGE` variable is externally configurable on ia32 so it should be possible to configure it in docker-wrapped build (and would probably be configurable for other targets like riscv64 when it would be CI-enabled).
> 
> Please note that CI explicitly sets `SYSPAGE` env variable so if you want to redo it locally 1:1, you should set it also:
> 
> https://github.com/phoenix-rtos/phoenix-rtos-project/blob/3cdb486e4276b9ac1fa8ff88e840c38d6939b665/.github/actions/phoenix-build/action.yml#L35-L41

Yeah, I think the best solution would be to leave passing SYSPAGE in this script and in our build action (in the future it may be needed), but we can remove setting `'psh pc-ata uart16550'` from workflows as it's not necessary. Do you approve such approach?

PR #942, opened 2023-12-01
imx6ull: rc.psh: mute supress

maska989:
## Description
JIRA: CI-393

Change cover situation when device has available Ethernet connection to startup `lwip`

## Motivation and Context
System init before suppress (With connected ):
Screenshot from 2023-11-27 10-08-20

System init after suppress:
Screenshot from 2023-11-27 10-08-02

## Types of changes

## How Has This Been Tested?

- [X] Already covered by automatic testing.
- [X] Tested by hand on: armv7a7-imx6ull-evk.

## Checklist:

- [X] All new and existing linter checks and tests passed.
- [X] My changes generate no new compilation warnings for any of the targets.

## Special treatment

### DIFF
--- a/_projects/armv7a7-imx6ull-evk/rootfs-overlay/etc/rc.psh
+++ b/_projects/armv7a7-imx6ull-evk/rootfs-overlay/etc/rc.psh
@@ -4,6 +4,8 @@ export HOME=/root
 W /bin/bind devfs /dev
 X /sbin/imx6ull-gpio
 W /sbin/dummyfs -m /var -D
+// suppressing dmesg message from lwip before execution
+X /bin/dmesg -D
 X /sbin/lwip enet:0x02188000:150:PHY:0.2:irq:-5:/dev/gpio5 enet:0x020b4000:152:no-mdio:PHY:0.1:irq:-6:/dev/gpio5
 X /bin/posixsrv
 X /bin/psh


nalajcie:
IMHO should not be part of the reference project (I would like to see the dmesg on console when I'm evaluating the firmware on the desk).

Also, exceptions while testing would be silenced then (not sure)?

I think it would be better to silence printing klog to console in `trunner` and reenable it (at the end of each test?) to see the possibly missed kernel logs?

damianloew:
> IMHO should not be part of the reference project (I would like to see the dmesg on console when I'm evaluating the firmware on the desk).
> 
> Also, exceptions while testing would be silenced then (not sure)?
> 
> I think it would be better to silence printing klog to console in `trunner` and reenable it (at the end of each test?) to see the possibly missed kernel logs?

Assuming that we want to silence logs for all of the tests maybe it would be good to silence it once before executing the whole campaign and enable it back at the end?

nalajcie:
> Assuming that we want to silence logs for all of the tests maybe it would be good to silence it once before executing the whole campaign and enable it back at the end?

If there would be any errors in dmesg, it would be helpful to have them assigned to the particular testcase they occurred in, not always the last testcase in the campaign. Obviously, approach like that needs some testing to verify it works as expected.

_projects/armv7a7-imx6ull-evk/rootfs-overlay/etc/rc.psh
```diff
@@ -4,6 +4,8 @@ export HOME=/root
 W /bin/bind devfs /dev
 X /sbin/imx6ull-gpio
 W /sbin/dummyfs -m /var -D
+// suppressing dmesg message from lwip before execution
```
gerard5, line 7: Use # for commented lines in psh scripts ([see](https://github.com/phoenix-rtos/phoenix-rtos-utils/blob/7b6b834b43cc1a5df57c10b5f9756cb973a9ba25/psh/pshapp/pshapp.c#L1115-L1116)).
```suggestion
# suppressing dmesg message from lwip before execution
```

PR #966, opened 2023-12-27
firmware_upg_imxrt117x

gorbanchovie:
## Description

Align offsets and enable using second, virtual memory bank 
## Motivation and Context

 It allows to run a phoenix based project from virtual bank (i.e. any memory offset) without modifying user script.   
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1176-nil.

## Checklist:

## Special treatment

### DIFF
--- a/_targets/build.common
+++ b/_targets/build.common
@@ -42,6 +42,11 @@ b_mkscript_user() {
 		fi
 	done
 
+	local RELATIVE_ALIAS=""
+	if [ -n "${NAME_USER2_SCRIPT}" ] && [ -n "${BANK2_BASE}" ]; then
+		RELATIVE_ALIAS=" -r"
+	fi
+
 	ksz=$((($(wc -c < "${PREFIX_PROG_STRIPPED}${KERNEL_FILE}") + SIZE_PAGE - 1) & PAGE_MASK))
 	poffs=$((KERNEL_OFFS + ksz))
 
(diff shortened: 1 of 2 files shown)


_targets/build.common
```diff
@@ -60,16 +63,16 @@ b_mkscript_user() {
 
 					PROGS+=("$name")
 					psz=$(wc -c < "${PREFIX_PROG_STRIPPED}$name")
-					printf "alias %s 0x%x 0x%x\n" "$name" "$poffs" "$psz"
+					printf "alias%s %s 0x%x 0x%x\n" "$RELATIVE_ALIAS" "$name" "$poffs" "$psz"
```
anglov, line 66: I would set RELATIVE_ALIAS to "-r" inside this function, not outside to avoid hacking possibilities :) 
Also you shouldn't rely on not setted variables in terms of purity (you can explicitly set it to empty string if not used)

Same applied to preinit

anglov (commented):
Generally ok, but I would put submodule update in separate PR

_targets/build.common
```diff
@@ -60,16 +65,16 @@ b_mkscript_user() {
...
```
agkaminski, line 68: Simplification possibility - you can use -r always, as default base is zero. I'm just pointing that out, I'm not sure it's a good idea. @anglov ?

agkaminski (approved):

_targets/build.common
gorbanchovie: I was also thinking about backward compatibility. Any project that doesn't upgrade plo on programming will fail due to lack of relative alias

_targets/build.common
```diff
@@ -42,6 +42,11 @@ b_mkscript_user() {
 		fi
 	done
 
+	RELATIVE_ALIAS=""
```
anglov, line 45: You could use `local` to not propagate this variable outside the function.

agkaminski (approved):

PR #990, opened 2024-01-30
[RTLD] Dynamic linker

badochov:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


badochov:
[CURRENT STATUS]
This changes will not address dynamic linking on NOMMU platforms. For those other mechanism will be needed.
  
Currently linking dynamically is only supported on ia32.   

Dynamic linker is adaptation of dynamic linker used by NetBSD.
Programs can be linked dynamically.
Libphoenix can be compiled as a shared libraries.
Syspage programs have to be compiled statically(most probably no way around it)

Locally I run nightly test campaign with all binaries linked dynamically and all tests passed but `busybox-cp` and `busybox-tr`, both failing with bogus log `I/O possible`.  It is weird as busybox uses different build system and is always linked statically.


TODO:
- Thread Local Storage support in dynamic libraries
- Support in kernel for 64bit platforms
- Support in dynamic linker for other platform than ia32. 
  - (To support other platforms asm parts of `ld.elf_so` have to be rewritten to support Phoenix process startup ABI.)
- Investigate busybox test fail 


Things to consider:
- I am not sure is dynamic linker can be compiled to support 32bit dynamic linking on 64bit platforms, whereas out kernel may (no idea if someone tested it)run 32bit binaries on 64bit

badochov:
Host tests fails due to devel image used being based on ubuntu 24 and having newer version of `libasan` than build image based on 22.04, won't fix

PR #1053, opened 2024-04-15
Fix .gitignore

agkaminski:
JIRA: RTOS-820

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
-.gitignore
 _fs/*
+!_fs/root-skel/
 _build/*
 _boot/*
 .DS_Store


.gitignore
```diff
@@ -1,5 +1,3 @@
-.gitignore
-_fs/*
```
nalajcie, line 2: `_fs/$TARGET` is being created/written to as a part of building process. 

We should probably move target rootfs to the `_buid` subdir but IMHO for now you need to whitelist only `rootfs-skel` subdir:

```
!_fs/root-skel/
```

nalajcie (approved):

PR #1157, opened 2024-09-09
feat(build): add VADDR_KERNEL variable

badochov:
JIRA: RTOS-908

## Description
On all platform but ARM, program headers are placed in loadable segment before `_init` section. Resulting from this  `_init` section start address isn't start address of first loadable segment. Due to that in PLO we may not be able to find proper segment for kernel if an address of the section that the kernel is to be loaded to = VADDR_KERNEL_INIT. Because of this additional linker script was required for sparc in kernel. 

On ARM we headers weren't loaded most probably because of a bug(or very fortunated accident) in our binutils patch.
THe [change](https://github.com/bminor/binutils-gdb/commit/a2bea3245ecf4ff67f923c85adf57aa78423eb45) introduced new file `armelf_phoenix.sh` but did't use it (`ld/configure.tgt`), instead it used armelf by default which defines `EMBEDDED` to true resulting in headers not being included.

Additionally this change splits `VADDR_KERNEL_INIT` to `ADDR_KERNEL_INIT` and `VADDR_KERNEL`. The latter is then passed to PLO and KERNEL thus avoiding redeclaration of it in kernel.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


badochov:
I'll create pr's in submodules if this approach is accepted

PR #1170, opened 2024-09-19
_targets: Add missing map for stm32l4x6

anglov:
Map contains area of OTP, ADC reference values and Option Bytes

## Motivation and Context

stm32l4-multi uses this map for ADC measurement, OTP read/write and Option Bytes read.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7m4-stm32l4x6-nucleo).

## Checklist:

## Special treatment

### DIFF
--- a/_targets/build.project.armv7m4-stm32l4x6
+++ b/_targets/build.project.armv7m4-stm32l4x6
@@ -61,6 +61,7 @@ PREINIT_SCRIPT=(
 	"map flash1 0x08080000 0x08100000 rxcb"
 	"map ram 0x20000000 0x20050000 rwxcb"
 	"map io 0x40000000 0x60000000 rw"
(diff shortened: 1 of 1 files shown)


anglov:
I'm not sure about naming.
About aligning - I choose wide area because we need all OTP (RW), Option Bytes(RO?) and single ADC registers(RO). Maybe you have a better idea, but I think it's good enough especially when only stm32l4-multi will access those.

Darchiv (approved):
LGTM. This map encloses reserved regions, but it seems to not be an issue, given that multidrv will have access to OTP and system memory anyway.

As for the name, maybe `sys`? But `otp` at least suggests it's not just RAM

agkaminski (approved):

PR #1240, opened 2024-11-26
[do not merge] rhealstone benchmark

lukileczo:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


_user/rhealstone/deadlock-break/main.c
```diff
@@ -0,0 +1,182 @@
...
+
+
+#define BENCHMARKS 1000
+
+#define ONE_TICK     26000 /* Number dependent on CPU. For loop to this count must be longer than sleep period (1 ms) */
+#define ONE_TICK_AVG 24990 /* For loop up to this count takes 1ms */
+
+bool deadBrk;
+atomic_int count = 0;
+atomic_bool done = false;
```
agkaminski, line 35: I think this should be `volatile`, but I'm not sure what exact effect it has on `atomic_`s

_user/rhealstone/deadlock-break/main.c
```diff
@@ -0,0 +1,182 @@
...
+
+	benchStart = getCntr();
+
+	priority(4);
+
+	usleep(0);
+
+	threadJoin(tid1, 0);
+
+	benchEnd = getCntr();
```
agkaminski, line 125: We introduce 2 syscalls overhead here that skews the result. Perhaps we could measure the time in `task1` instead?

_user/rhealstone/irq-latency/main.c
```diff
@@ -0,0 +1,79 @@
...
+#include <sys/time.h>
+#include <sys/threads.h>
+
+#include <board_config.h>
+#include <rhcommon.h>
+
+
+#define BENCHMARKS 10000
+
+atomic_uint_least64_t benchEnd = 0;
```
agkaminski, line 32: Atomic might give unwanted overhead, especially 64 bit one - we trigger the interrupt via SW, so should be fine to use plain `volatile`

_user/rhealstone/irq-latency/main.c
```diff
@@ -0,0 +1,79 @@
...
+#ifdef __CPU_GR740
+#define IRQ_UNUSED 13
+#endif
+
+uint64_t results[BENCHMARKS] = { 0 };
+
+
+int irqHandler(unsigned int n, void *arg)
+{
+	if (n == IRQ_UNUSED) {
```
agkaminski, line 43: This `if` is not strictly needed - we provide `n` argument as a way for one handler to handle more than one interrupt, check on `n` is not enforced

_user/rhealstone/irq-latency/main.c
```diff
@@ -0,0 +1,79 @@
...
+
+	return 0;
+}
+
+
+int main(void)
+{
+	BENCH_NAME("Interrupt latency");
+
+	uint32_t *irqCtrl = mmap(NULL, _PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_DEVICE | MAP_PHYSMEM | MAP_ANONYMOUS, -1, (uintptr_t)INT_CTRL_BASE);
```
agkaminski, line 55: Should be volatile, actually it's kinda weird that the access to the `irqCtrl` hasn't been optimized out. Are we sure we have `-O2`?

_user/rhealstone/msg-latency/main.c
```diff
@@ -0,0 +1,185 @@
...
+		return -1;
+	}
+
+	return 0;
+}
+
+
+void queueDestroy(struct queueCtx *ctx)
+{
+	free(ctx->buffer);
```
agkaminski, line 76: Use after free

_user/rhealstone/msg-latency/main.c
```diff
@@ -0,0 +1,185 @@
...
+
+	for (volatile int i = 0; i < BENCHMARKS; i++) {
+	}
+
+	for (volatile int i = 0; i < BENCHMARKS; i++) {
+	}
+
+	loopOverhead = getCntr() - loopOverhead;
+
+	uint64_t joinOverhead = threadJoinOverhead();
```
agkaminski, line 159: for performance calibration, I assume. `nop` version might be better, we use non-volatile iterator in the use-case

agkaminski (commented):
General remark - watch out for atomics, these might have some heavy overheads (including unexpected mutex in some cases). Perhaps it's ok, but we can't be sure without inspecting resulting binary

Darchiv:
I think phoenix-rtos-utils/benchmarks/rhealstone will be a better place for this. @agkaminski What do you think?

agkaminski:
> I think phoenix-rtos-utils/benchmarks/rhealstone will be a better place for this. @agkaminski What do you think?

Sounds good

Darchiv:
Moved to https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/236

PR #1257, opened 2025-01-17
Adamdebek/lsb vsx

adamdebek:
## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

- [x] Tested by hand on: (list targets here).

## Checklist:

## Special treatment

- [x] I will merge this PR by myself when appropriate.

### DIFF
--- a/_targets/ia32/generic/build.project
+++ b/_targets/ia32/generic/build.project
@@ -33,6 +33,7 @@ export PORTS_OPENVPN=y
 export PORTS_JANSSON=y
 export PORTS_CURL=y
 export PORTS_MICROPYTHON=y
+export PORTS_LSB_VSX=y
 
 #
 # Platform dependent parameters
--- a/_targets/ia32/generic/nvm.yaml
+++ b/_targets/ia32/generic/nvm.yaml
@@ -1,5 +1,5 @@
 hd0:
-  size: 0x4000000 # 64 MB
+  size: 0x8000000 # 128 MB
   block_size: 512
   partitions:
     - name: plo


_targets/ia32/generic/build.project
```diff
@@ -33,6 +33,7 @@ export PORTS_OPENVPN=y
 export PORTS_JANSSON=y
 export PORTS_CURL=y
 export PORTS_MICROPYTHON=y
+export PORTS_LSB_VSX=y
```
damianloew, line 36: Let's place it in _projects/ia32-generic-qemu as you didn't test it on `ia32-generic-pc`

_targets/ia32/generic/nvm.yaml
```diff
@@ -1,5 +1,5 @@
 hd0:
-  size: 0x4000000 # 64 MB
+  size: 0x8000000 # 128 MB
```
damianloew, line 2: @Darchiv can we increase the size?

damianloew (changes requested):

_targets/ia32/generic/build.project
adamdebek: Was at least one port tested on ia32-generic-pc?

I think keeping it in 1 place is more convenient and error prone

damianloew: I am not sure, but if not - they shouldn't be built in my opinion. However, I see that some other new ports like `lua` are placed in `_targets`, so I am leaving the decision to @Darchiv.

_targets/ia32/generic/nvm.yaml
Darchiv: Seems fine.

@adamgreloch did not encounter any issues as far as I knoiw

_targets/ia32/generic/build.project
Darchiv: Well, most of the ports are put in `_targets`, as they mainly depend on the hardware architecture itself and not low-level platform details (peripherals, etc.). I think it is fine to put it in `_targets` now.

FYI: We may need to improve how ports are built as a whole and this use case should also be considered.

damianloew (approved):

Darchiv:
This PR is already merged, but please change its name to something relevant, so as to make other users' and contributors' life easier

PR #1262, opened 2025-01-22
lighttpd: lighttpd.conf updates

gstrauss:
lighttpd: lighttpd.conf updates

## Description
- prefer lighttpd defaults
- resource tuning
- remove mod_evasive and mod_usertrack (removed from upstream)
- comment out modules not otherwise used in lighttpd.conf

## Motivation and Context
modern lighttpd with bug fixes and HTTP/2 (optional)
fixes issue blocking https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/90

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/_fs/root-skel/etc/lighttpd.conf
+++ b/_fs/root-skel/etc/lighttpd.conf
@@ -16,6 +16,7 @@ var.conf_dir    = "/etc/lighttpd"
 var.http_port  =   80
 var.https_port = 443
 
+## XXX: debug logging not recommended for production; may be noisy
 debug.log-file-not-found = "enable"
 #debug.log-request-handling = "enable"
 #debug.log-request-header = "enable"
@@ -29,20 +30,18 @@ server.port = http_port
 ## enabled modules
 server.modules = (
   "mod_indexfile",
-  "mod_dirlisting",
-  "mod_staticfile",
   "mod_access",
   "mod_accesslog",
   "mod_alias",
-  "mod_auth",
-  "mod_authn_file",
+#  "mod_auth",
+#  "mod_authn_file",
   "mod_openssl",
-#  "mod_evasive",
-  "mod_redirect",
+#  "mod_redirect",
   "mod_rewrite",
 #  "mod_setenv",
-#  "mod_usertrack",
   "mod_cgi"
+  "mod_dirlisting",
+  "mod_staticfile",
 )
 
 ## Document root
@@ -62,7 +61,7 @@ server.pid-file = state_dir + "/lighttpd.pid"
 # we can still serve static files and output of the CGI scripts. The only
 # exception is the '/log' subdir, where we want to get 404 error, if we're trying
 # to download log file that does not exist.
-url.rewrite-repeat-if-not-file = ( "^/(?!log/)(.*)" => "/index.html" )
+url.rewrite-repeat-if-not-file = ( "^/(?!log/)" => "/index.html" )
 
 #######################################################################
 ##  Logging Options
@@ -85,28 +84,28 @@ accesslog.format = "%h %l %u %t \"%r\" %b %>s \"%{user-agent}i\" \"%{referer}i\"
 
 #######################################################################
 ##  Tuning/Performance
+## https://wiki.lighttpd.net/Docs_ResourceTuning
+## https://wiki.lighttpd.net/Docs_Performance
 #######################################################################
 server.event-handler = "poll"
-server.network-backend = "write"
-server.max-fds = 64
-server.stat-cache-engine = "simple"
+server.max-fds = 128
 
 ## performance options
+## (too low might lower performance due to client having to reconnect;
+##  too high might deny other clients if server.max-connections is reached)
 server.max-keep-alive-requests = 6
-server.max-keep-alive-idle = 15
-server.max-read-idle     = 60
-server.max-write-idle    = 360
 
 ## maximum concurrent connections the server will accept (1/2 of server.max-fds)
 server.max-connections = 16
 
-## 32 MB in kB, only applies to POST (header + body)
+## 32 MB in kB, max size of request header + request body + request trailers
 server.max-request-size = 32768
 
 #######################################################################
 ##  Filename/File handling
 #######################################################################
 
+## (XXX: prefer a shorter list)
 index-file.names += (
   "index.xhtml", "index.html", "index.htm", "default.htm", "index.cgi", "login.html"
 )
@@ -116,9 +115,6 @@ url.access-deny             = ( "~", ".inc" )
 
 ## which extensions should not be handle via static-file transfer
 static-file.exclude-extensions = ( ".cgi")
-
-## Should lighttpd follow symlinks?
-server.follow-symlink = "enable"
 
 ## defaults to /var/tmp as we assume it is a local harddisk
 server.upload-dirs = ( "/var/tmp" )
@@ -160,6 +156,8 @@ $HTTP["url"] =~ "/cgi-bin/" {
 #######################################################################
 
 # use HTTPS
+## (XXX: use better syntax available with modern lighttpd releases)
+## https://wiki.lighttpd.net/Docs_SSL
 $SERVER["socket"] == ":" + https_port {
     ssl.engine                  = "enable"
 #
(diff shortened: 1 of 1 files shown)


gstrauss:
ping @Darchiv

_fs/root-skel/etc/lighttpd.conf
```diff
@@ -29,20 +30,18 @@ server.port = http_port
...
   "mod_accesslog",
   "mod_alias",
-  "mod_auth",
-  "mod_authn_file",
+#  "mod_auth",
+#  "mod_authn_file",
   "mod_openssl",
-#  "mod_evasive",
-  "mod_redirect",
+#  "mod_redirect",
```
Darchiv, line 39: Maybe modules in a generic/default configuration file should not be commented-out, so that CI will always build them (and verify they really build)? I am aware that there are plenty of modules and some of them may conflict/overlap in functionality and require additional packages (e.g. mod_gnutls, mod_mbedtls, mod_openssl, mod_wolfssl), in which case we wouldn't want to enable them, but maybe there should be a set of most used/generic enough always-built modules.

This approach assumes that modules are grepped from the configuration file into plugin-static.h

_fs/root-skel/etc/lighttpd.conf
```diff
@@ -16,6 +16,7 @@ var.conf_dir    = "/etc/lighttpd"
 var.http_port  =   80
```
Darchiv, line 16: If you aim to make this configuration file more generic ("prefer lighttpd defaults") then maybe change the

> lighttpd single-file conf for Current DCU

line from the header to something less specific.

_fs/root-skel/etc/lighttpd.conf
```diff
@@ -62,7 +61,7 @@ server.pid-file = state_dir + "/lighttpd.pid"
 # we can still serve static files and output of the CGI scripts. The only
 # exception is the '/log' subdir, where we want to get 404 error, if we're trying
 # to download log file that does not exist.
-url.rewrite-repeat-if-not-file = ( "^/(?!log/)(.*)" => "/index.html" )
+url.rewrite-repeat-if-not-file = ( "^/(?!log/)" => "/index.html" )
```
Darchiv, line 64: Is the `(.*)` part removed due to changes in the latest version of lighttpd?

This line should probably be removed from a generic configuration as it is specific for Single Page Applications. Maybe there should be a separate configuration file for such applications (for reference purposes) or left as a comment?

_fs/root-skel/etc/lighttpd.conf
```diff
@@ -85,28 +84,28 @@ accesslog.format = "%h %l %u %t \"%r\" %b %>s \"%{user-agent}i\" \"%{referer}i\"
 
 #######################################################################
 ##  Tuning/Performance
+## https://wiki.lighttpd.net/Docs_ResourceTuning
+## https://wiki.lighttpd.net/Docs_Performance
 #######################################################################
 server.event-handler = "poll"
-server.network-backend = "write"
```
Darchiv, line 90: Maybe leave these as a comment? Low-level options may be very useful in the context of embedded systems. Have you checked that lighttpd's implicit choice works on Phoenix-RTOS?

Darchiv (commented):
I have left some remarks.

I appreciate that you make the "default" configuration file (as _root-skel_ directory itself suggests) more default. It should probably be moved somewhere else (port's directory in phoenix-rtos-ports?), but there is also the possibility that a per-target configuration is needed for lighttpd to run at all, which would be stored in _phoenix-rtos-project/_projects/*/rootfs-overlay/etc_, but that's not in scope of this PR. Let's focus on upgrading lighttpd to 1.4.77 and make further improvements when it's done :)

_fs/root-skel/etc/lighttpd.conf
gstrauss: This is a modification to remove excess, which can make a difference on embedded systems.  This is unrelated to any change in lighttpd.

Regex captures use more resources and can be abused in regex attacks (security).  In this case, there is no use of the regex capture, and therefore it can be safely removed without changing the intended semantics of the regex, which is a prefix match using a zero-width negative lookahead assertion (which is non-capturing).

_fs/root-skel/etc/lighttpd.conf
gstrauss: The initial thrust of this PR is to unblock https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/90

Please do as you sit fit to keep this moving forwards.

I have expressed my opinions on the current choices of grep and overload of `server.modules` for both generating `plugin-static.h` for the static build and also enabling the modules at runtime.  I have made alternative suggestions, but I think that should be for a future discussion.

For this PR, the `"mod_evasive"` and `"mod_usertrack"` must be removed, as they no longer exist in lighttpd.

If you want the other lines to remain uncommented, please indicate so.  Alternatively, you have permission to edit this PR, so you can avoid unnecessary latency of asking me to make changes by making any changes you like to this PR to keep it moving forwards.

_fs/root-skel/etc/lighttpd.conf
gstrauss: What do you suggest?  This still meets the label that (someone) in phoenix-rtos put at the top of the file.

The official lighttpd source code in lighttpd 1.4.77 changes the sample configuration file format to include a `lighttpd.conf` "header" file with a basic intro of the configuration, and subsequent `lighttpd.annotated.conf` with more knobs listed.
https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/master/doc/config/lighttpd.conf
https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/master/doc/config/lighttpd.annotated.conf

That was done for compatibility with current use.

Personally, I think the historical "list all the configuration options commented out" is not the best approach when most people would do better to leave most of the defaults alone.  lighttpd is very modular and can be complex, and all of those extra options usually result in people not reading documentation, and being overwhelmed by the options.

On the other hand, for a simple program, a small configuration file with a small set of tunable options -- especially when the user is expected to make choices and change -- may be more appropriate.  I do not think that is applicable to the volume of potential options in lighttpd, especially on an embedded system with a static build of lighttpd which omits lighttpd modules and therefore omits entire lists of options.

https://wiki.lighttpd.net/Docs_ConfigurationOptions

_fs/root-skel/etc/lighttpd.conf
gstrauss: The comments is the links to the wiki with some explanations on the tunables and how they might be used and in what situations.  I personally disagree with leaving these commented-out, especially when "write" is the least efficient and should not be used unless others are not available.

However, I have made my suggestion and provided my justifications.  Please make your choice on what you will accept and edit this PR as you see fit so that you will accept it.

gstrauss: > Have you checked that lighttpd's implicit choice works on Phoenix-RTOS?

lighttpd will use `sendfile()` if available and detected during compilation.  At runtime, if `sendfile()` fails (such as `ENOTSUP` or `EOPNOTSUPP` for a specific filesystem), then lighttpd will transparently fall back to `writev()`.  `write()` is for Windows lusers.

gstrauss:
@Darchiv ping.

gstrauss:
@Darchiv, @agkaminski  ping.

It has been over a month since @Darchiv provided feedback and I responded the same day to answer questions.  Is your intention to discourage contributors?

For this PR, the "mod_evasive" and "mod_usertrack" must be removed, as they no longer exist in lighttpd.  If you feel unable to accept this PR because you want to change larger things out-of-scope, then please narrow the scope of this PR to remove "mod_evasive" and "mod_usertrack" to unblock https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/90

agkaminski:
@gstrauss Sorry! We're having a very busy time right now. We **do intend** to merge the PR, I hope we'll be able to do this upcoming week.

gstrauss:
> @gstrauss Sorry! We're having a very busy time right now. We **do intend** to merge the PR, I hope we'll be able to do this upcoming week.

Thank you.  I was not trying to rush you; I was merely looking for a response and (reasonable) estimate for when feedback might be provided.  A week is fine.  So are two weeks if you communicate that expectation.

_fs/root-skel/etc/lighttpd.conf
Darchiv: Ok, so it is redundant. That makes sense

_fs/root-skel/etc/lighttpd.conf
Darchiv: If functions are detected at build-time and results checked at runtime, then removing this line should be fine. We aim to fix any issues regarding use of `sendfile()`/`writev()`/`write()` should they arise.

_fs/root-skel/etc/lighttpd.conf
Darchiv: Ok, let's focus on removing `"mod_evasive"` and `"mod_usertrack"`

PR #1285, opened 2025-03-31
scripts: add openocd configs

dsabala:
Added a few openocd config files for boards supported by Phoenix RTOS

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/scripts/openocd/ftdi-te0790.cfg
+++ b/scripts/openocd/ftdi-te0790.cfg
@@ -1,0 +1,9 @@
+# This is OpenOCD configuration for TE0790-03 Trenz Electronic XMOD FTDI JTAG
+# https://www.digikey.pl/pl/products/detail/trenz-electronic-gmbh/TE0790-03/10071026
+
+adapter driver ftdi
+ftdi vid_pid 0x0403 0x6010
+ftdi channel 0
+ftdi layout_init 0x0008 0x000b
+transport select jtag
+adapter speed 12000
--- a/scripts/openocd/ftdi-zcu104.cfg
+++ b/scripts/openocd/ftdi-zcu104.cfg
@@ -1,0 +1,11 @@
+# This is OpenOCD configuration for EK-U1-ZCU104-G on board JTAG debugger
+# https://www.digikey.pl/pl/products/detail/amd/EK-U1-ZCU104-G/9380242
+
+adapter driver ftdi
+ftdi vid_pid 0x0403 0x6011
+ftdi channel 0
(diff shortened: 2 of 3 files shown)


agkaminski (dismissed):

Darchiv:
> @nalajcie @Darchiv What do you think, do these scripts fit here or should be moved to the phoenix-rtos-build?

Hard to tell. In phoenix-rtos-project we have per-device/per-product definitions of memory layout, partitions, and syspage programs (declarative, e.g. nvm.yaml) and build.project with "how to build an image" procedures (imperative). OpenOCD config files may be useful when flashing/debugging and they usually rely on the mentioned target device definitions (flash addr etc.) so it makes sense to put them close to each other, potentially even the same _target directory. (Maybe there should be `b_flash_target()` function in every build.project to abstract away some parameters or make flashing a one-step process for simple cases?)

In phoenix-rtos-build, on the other hand, we have such scripts already. The scripts from this PR seem not to be a part of some abstraction or public build pipeline, so I think they can be put into p-r-build for now to be consistent with the current approach.

General remark - please don't use phrases like "few files" or "few fixes". There are just three boards you provide scripts for and they can be listed in the descripton directly. Users and developers will find it easier to track and find relevant changes should they need to browse history.

dsabala:
Maybe we should place those OpenOCD configuration files and other helper tools in the phoenix-project-pilot repository? Once they mature, we can consider moving them here.

agkaminski:
> Maybe we should place those OpenOCD configuration files and other helper tools in the phoenix-project-pilot repository? Once they mature, we can consider moving them here.

I am almost certain that this way these will never make their way to github ;)

dsabala:
@agkaminski Maybe we should move those scripts to phoenix-rtos-build/scripts, as this would allow us to add a Python script to assist with uploading flash images on ZynqMp. This script could leverage other Python modules already located in the mentioned directory to gain information about memory partitioning. It could help reduce the time required to upload flash memory and lower the cognitive load for developers who perform this task multiple times a day

agkaminski:
> @agkaminski Maybe we should move those scripts to phoenix-rtos-build/scripts, as this would allow us to add a Python script to assist with uploading flash images on ZynqMp. This script could leverage other Python modules already located in the mentioned directory to gain information about memory partitioning. It could help reduce the time required to upload flash memory and lower the cognitive load for developers who perform this task multiple times a day

Yes, I think that phoenix-rtos-build is the best place for them

dsabala:
Those files and other help scripts better suits phoenix-rtos-build

PR #1381, opened 2025-08-21
ci: fix in-docker paths

nalajcie:
## Description
When running docker manually - we need to reflect the host paths (no Path translation is done in the runner).

This should fix the annotations (problem-matchers file-match).

## Motivation and Context

JIRA: CI-582

## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/.github/actions/phoenix-runner/action.yml
+++ b/.github/actions/phoenix-runner/action.yml
@@ -42,14 +42,17 @@ runs:
       # runner args:
       #  -v  # output SKIP subresults at a top level
       #  -Oresults.csv # creates also junit xml
+      # when running docker manually - we need to reflect the host paths (no Path translation is done in the runner)
       docker run --rm --privileged \
-        -v "${{ github.workspace }}:/workspace" \
-        -w /workspace \
-        -v "${HOME}:/github/home" \
-        -v "${RUNNER_TEMP}:/github/workflow" \
-        -v "${GITHUB_ENV}:/github/file_commands/env" \
-        -v "${GITHUB_OUTPUT}:/github/file_commands/output" \
+        -v "${{ github.workspace }}:${{ github.workspace }}" \
+        -w "${{ github.workspace }}" \
+        -v "${HOME}:${HOME}" \
+        -v "${RUNNER_TEMP}:${RUNNER_TEMP}" \
+        -v "${GITHUB_ENV}:${GITHUB_ENV}" \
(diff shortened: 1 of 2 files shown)


.github/actions/phoenix-runner/action.yml
```diff
@@ -42,14 +42,17 @@ runs:
...
-        -v "${HOME}:/github/home" \
-        -v "${RUNNER_TEMP}:/github/workflow" \
-        -v "${GITHUB_ENV}:/github/file_commands/env" \
-        -v "${GITHUB_OUTPUT}:/github/file_commands/output" \
+        -v "${{ github.workspace }}:${{ github.workspace }}" \
+        -w "${{ github.workspace }}" \
+        -v "${HOME}:${HOME}" \
+        -v "${RUNNER_TEMP}:${RUNNER_TEMP}" \
+        -v "${GITHUB_ENV}:${GITHUB_ENV}" \
+        -v "${GITHUB_OUTPUT}:${GITHUB_OUTPUT}" \
```
damianloew, line 52: Is   
```
 -v "${{ github.workspace }}:${{ github.workspace }}" \
 -v "${RUNNER_TEMP}:${RUNNER_TEMP}" \
 -v "${GITHUB_ENV}:${GITHUB_ENV}" \
 -v "${GITHUB_OUTPUT}:${GITHUB_OUTPUT}" \
 ```

needed when we are mounting the whole home?

I am not 100% sure that it's working this way, but I suppose that HOME is `/home/runner`, RUNNER_TEMP is `/home/runner/work/_temp`, github.workspace is `/home/runner/work/repo/repo`, so then maybe directories inside home doesn't have to be mounted (?)

nalajcie: I think this is the implementation detail of runner (on linux, on other OSes the paths might be different)  - and is subject to change in the future. I wanted to stay on the safe side and replicate the mounts being done when the runner is doing the container action run (we're not needing half of them right now) - see the source rof container action handler here: https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/ContainerActionHandler.cs#L194

damianloew: Understand, I agree that safer approach is better

damianloew: Idk how important is it (and whether it's used further), but maybe `How Has This Been Tested?` and `Checklist:` is worth completing

PR #1534, opened 2026-01-14
 _targets/stm32n6: use stm32n6_sign_fsbl.py for signing plo image

jmaksymowicz:
## Description
Previously our build script used `STM32_SigningTool_CLI` for signing the PLO image so it can boot on STM32N6. Unfortunately, this tool is proprietary so we don't have control over it and integrating it into our CI pipeline would be difficult. Also, version 2.21.0 breaks our build process, as it generates less padding than our application assumes.

We have our own open-source script that is now fully functional and can be used instead. The dependencies of this new script are stored in `phoenix-rtos-build/scripts/stm32n6_requirements.txt`.

## Motivation and Context
Fix RTOS-1199

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6-nucleo

## Checklist:

## Special treatment

### DIFF
--- a/_targets/armv8m55/stm32n6/build.project
+++ b/_targets/armv8m55/stm32n6/build.project
@@ -11,9 +11,6 @@
 
 CROSS=arm-phoenix-
 
-# Path to STM32_SigningTool_CLI. User may wish to set a custom path.
-STM32SIGNTOOL=${STM32SIGNTOOL:="STM32_SigningTool_CLI"}
-
 export BUSYBOX_CONFIG="${PROJECT_PATH}/busybox_config"
 export DROPBEAR_CUSTOM_CFLAGS="-DLTC_NO_BSWAP"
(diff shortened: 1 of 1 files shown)


damianloew:
- additional packages in `build`: As discussed with @wmlynik, after adding this change it won't be needed
 - extending docs: From what I know not needed too, but @wmlynik could you confirm that?

nalajcie:
When I try to run the script in current `build`  image:
image

Please explain how the change in `build` image is not needed?

wmlynik:
After adding this change `STM32_SigningTool_CLI` won't be needed for building, but we need to add dependencies of the script to the `build` docker image.
The docs regarding the dev env setup  should also be extended with installation of those dependencies in case of native toolchain.

Darchiv (approved):

PR #1623, opened 2026-05-11
VERSION: track next version tag

oI0ck:
port_manager gets version candidates from git describe. This is correct for already existing ports, but not for newly ported ports, which may require system changes in order to work. They might need to require a newer RTOS version, for which there is no tag yet.

VERSION metafile tracks the next release tag, so that it can be added as a constraint to port_manager.

At all times, this file should contain either:
 - If checked out on master or a release branch, it should contain next release version from its base branch.
 - If checked out on a tag, it should contain said tag.

This file should be bumped every release on master or release branches.

JIRA: RTOS-1318

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
- https://github.com/phoenix-rtos/phoenix-rtos-build/pull/278
- [X] I will merge this PR by myself when appropriate.


adamgreloch (dismissed):

VERSION
```diff
@@ -0,0 +1 @@
+v3.4.0
```
nalajcie, line 1: version is eg. `3.4.0` and is only tagged in git using `v` prefix (so the full tag is `v3.4.0`).

I don't like the disassociation between kernel version and "project" version, but I'm not sure how to solve it gracefully. 

Also: not a fan of setting the version to "3.4.0" only (right now it's harmless, but soon people would start to use it in other context and soon we will have 3.4.0 spread out across binaries as "current" version.

I've seen two approaches regarding versioning:
 - after `3.3.0` release we bump the current version to `3.3.99` (patch version 99 means "next")
 - using suffix like `3.3.0-next`, `3.4.0-rc`, etc.

adamgreloch: I'm in for `3.4.0-rc` as it seems the cleanest from port requirements perspective. It will satisfy `>=3.4.0` requirement just like `3.4.0` and still be clear that it is not an actual release.

`3.4 - eps` or `3.3.0-next` would require rewriting port defs once the 3.4 is released

oI0ck: @nalajcie 
From my point of view, the project version propagates to other parts of the system as they are often dependant on eachother.

Isn't 3.4.0 what we are planning to release? Release branches track older versions, hence, their VERSION file would be different.

Not using the "+1" version from latest release would only give us freedom to pick the version freely, IF we ever decide to break from this convention.

Using 3.4 here is convenient, as we don't have to change new ports from "3.3-current" to "3.4" on release.

nalajcie: What I'm saying is that somebody will soon use this file as eg. `cat VERSION > /etc/os-version` or in `phoenix-rtos-kernel` as the current version - when indeed it's not a 3.4.0 yet. 

Semantically `3.3.99` or `3.4.0-rc` is better and (if I understand correctly) will cover ports versioning (future 3.4.0 is >= 3.3.99 (version in git), future 3.4.0 is >= 3.4.0-rc (version in git)).

Both are common practices, for example see:
 - qemu (bumps 10.2.0 immediately to 10.2.50, RC versions are 10.2.9x (91 for rc1, 92 for rc2, etc.) - https://github.com/qemu/qemu/commits/master/VERSION
 - linux actually keeps release version 7.0.0 for several weeks until  7.1.0-rc1 release (but the buildsystem checks if you're at the tagged commit - if not, automatically appends suffix)
 - gcc uses `gcc/BASE-VER` and `gcc/DEV-PHASE` (keeps last-released version in BASE-VER, changes DEV-PHASE, bumps BASE-VER only on new release)
 - LLVM bumps version in master branch to X.Y.Z-git, never bumps to X.Y.Z only (https://github.com/llvm/llvm-project/commits/main/cmake/Modules/LLVMVersion.cmake)
 
 
 IMHO we should describe the versioning process somewhere (after we would agree on convention).

oI0ck: Agreed, maybe `3.4.0-current`, following FreeBSD convention?

oI0ck:
Do we want to track kernel version using this file (That would require a change to kernels `Makefile`)?

oI0ck:
So, I had some thoughts on this matter over the weekend.

First of all, I think I'd still prefer 3.4.0 over 3.4.0-whatever in this file. This comes down mainly to what one have to do when creating a new release.

If we use 3.4.0, (next release version in VERSION), then for every release we'd have to do the following:
1. Create a release branch
2. Tag on latest commit
3. Bump VERSION on both master and release

On the other hand, this is what we'd have to do when keeping versions with suffixes:
1. Create a release branch
2. Commit removing the suffix from VERSION file
3. Tag this commit
4. Bump version on master and on release (and add the suffix)

What pains me, is that keeping the suffix, just to signal that this is not the defacto version, causes a worse and more annoying release cycle, because we'd have to have a special commit in the release branch that would delete the suffix to create a revision of VERSION that contains the tags name, and then we'd have to add it again.

On the other hand, we can rename the VERSION file to TRACKING or NEXT, such that it is not obviously the current version.

Another matter is versioning other components, like PLO or the kernel. As of now, the kernel is versioned using tags and a dummy version, bumped every release, used if git is unavailable during build time. If we'd like to unify versioning along with this change, we'd have to agree on what information do we want to output with our version string.
Namely, the kernel uses now `git describe` to get the latest tag, current revision, and commits from said tag.

Assuming that we'd want to use VERSION to version the kernel, we'd have to remove the "offset from tag" information, as this would add confusion, or add `git describe` output as a separate source of information.

RFC? @nalajcie @Darchiv

nalajcie:
-  don't optimize a process we're doing once a year (and could easily be reduced to shell script or CI job)
- `VERSION` is a bad name, at least `OS-VERSION` or something like that
- think about a customer (e.g. you're releasing a SDK for a product and will have `/OS-VERSION` in the package which clearly tells a lie - the OS version is not 3.4.0)

PR #1755, opened 2026-09-10
_projects/armv8m55-stm32n6-nucleo: enable RTT for flashing in plo

jmaksymowicz:
Enable RTT in `board_config.h` and update preinit script accordingly.
Add OpenOCD script for writing plo into RAM and activating RTT.

## Description

On STM32N6 there was a long-standing problem of how to write a system image to the board in a user-friendly manner and without using proprietary tools. This PR enables a flashing workflow like this:
1. Put board into development boot mode
2. Run OpenOCD script to write plo into memory and open RTT communication.
3. Run phoenixd, which will exchange data from host over RTT
4. Issue commands to plo to erase Flash and copy disk image from host to Flash

This method of flashing the board is reasonably performant (copying data works at about 28 KiB/s), relatively user-friendly, uses only Phoenix-developed tools and is similar to the solutions on other platforms.

This method is the is intended for use by beginners working on Nucleo boards with secure boot disabled. For boards other than Nucleo (which don't allow easy access to SWD interface) or when secure boot is enabled, other methods need to be used.

UPDATE: Also added script for flashing via RAM-disk. In testing it achieves speeds of over 160 KiB/s, at this point it is bottlenecked by Flash erase speed. Flashing could be further optimized by supporting multiple erase sizes in plo, but even maximum theoretical speed is somewhere around 220 KiB/s so not much higher.

## Motivation and Context
YT: RTOS-1177

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6-nucleo

## Checklist:

## Special treatment

### DIFF
--- a/scripts/stm32n6-plo-rtt.cfg
+++ b/scripts/stm32n6-plo-rtt.cfg
@@ -1,0 +1,48 @@
+if { [info exists PLO_PATH] == 0 } {
+	# Try to get from environment variables
+	if { [dict exists $env PLO_PATH] } {
+		set PLO_PATH [dict get $env PLO_PATH]
+	} else {
+		puts "ERROR: PLO_PATH not set. Please set PLO_PATH to point at plo-ram.elf file."
+		shutdown
+	}
+}
+
+init
+reset halt
+
+set x 0
+set max_attempts 2
+while { $x < $max_attempts } {
+	load_image $PLO_PATH
+	# Extract reset vector from the vector table and jump to it
+	resume [mrw 0x34180404]
+	# Wait a short time for plo and RTT to be initialized
+	sleep 500
+	halt
+	# On cold boot soon after resume the chip will reset and go back into Boot ROM.
+	# Check if we're executing out of Boot ROM now.
+	if { [lindex [get_reg {pc}] 1] < 0x20000000 } {
+		# After a reset something in the debug probe doesn't get reinitialized properly.
+		# For some inscrutable reason performing a byte or halfword (but not word!) access
+		# to any address makes the interface work again.
+		mdb 0x34180000
+	} else {
+		break
+	}
+
+	incr x
+}
+
+if { $x == $max_attempts } {
+	puts "ERROR: Failed to run plo"
+	shutdown
+}
+
+resume
+
+rtt setup 0x3001ff00 0x100 "RTT PIPE"
+rtt server start 18021 0
+rtt server start 18022 1
+rtt polling_interval 1
+rtt start
--- a/_projects/armv8m55-stm32n6-nucleo/board_config.h
+++ b/_projects/armv8m55-stm32n6-nucleo/board_config.h
@@ -20,6 +20,8 @@
 #define UART_PIN_TX  5
 #define UART_PIN_RX  6
 
+#define RTT_ENABLED_PLO 1
+
 #define XSPI2           1
 #define XSPI2_CLOCK_DIV 1 /* 200 MHz */
 #define XSPI2_STORAGE   1
--- a/_targets/armv8m55/stm32n6/preinit.plo.yaml
+++ b/_targets/armv8m55/stm32n6/preinit.plo.yaml
@@ -1,11 +1,12 @@
-size: 0x200
+size: 0x400
 is_relative: False
 
 contents:
   # NOTE: All data/code maps need to be cacheable to allow unaligned accesses
   # env.BOOT_DEVICE should correspond to one of the map names
   - map itcm       0x10000000 0x10020000 rxcb  # ITCM
-  - map dtcm       0x30000000 0x30020000 rwcb  # DTCM
+  - map dtcm       0x30000000 0x3001ef00 rwcb  # DTCM
+  - map rtt        0x3001ef00 0x30020000 rw    # RTT buffers
   - map axi_kernel 0x34014000 0x34100000 cb    # Kernel data
   # Note: PLO is loaded by BootROM at 0x34180000, inside axi_app
   - map axi_app    0x34100000 0x343c0000 rwxcb # Applications code/data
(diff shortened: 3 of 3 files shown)


scripts/stm32n6-plo-rtt.cfg
```diff
@@ -0,0 +1,38 @@
+init
+reset halt
+
+set x 0
+set max_attempts 2
```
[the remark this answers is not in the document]
jmaksymowicz: Done. Also added option to give path through environment variables.

scripts/stm32n6-plo-rtt.cfg
```diff
@@ -0,0 +1,38 @@
...
+}
+
+if { $x == $max_attempts } {
+	puts "ERROR: Failed to run plo"
+	shutdown
+}
+
+resume
+
+rtt setup 0x3001ff00 0x100 "RTT PIPE"
```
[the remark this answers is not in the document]
jmaksymowicz: I don't think this is necessary. Our implementations of RTT always put the control block in the last 256 bytes of its designated memory map. Widening the search space like this would also result in slower initialization.

scripts/stm32n6_ramdisk_flash.py
```diff
@@ -0,0 +1,292 @@
...
+    PLO_CMD_TIMEOUT = 30
+
+    def __init__(self, scripts_dir: Path, boot_dir: Path, adapter_speed: Optional[int] = None) -> None:
+        # Nice to be configurable
+        self.ramdisk_name = "ramdisk"
+        self.ramdisk_addr = 0x34100000
+        self.ramdisk_size = 0x00080000
+
+        self.boot_dir = boot_dir.absolute()
+        plo_path = self.boot_dir / self.PLO_FILENAME
```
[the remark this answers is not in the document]
jmaksymowicz: Done

scripts/stm32n6_ramdisk_flash.py
```diff
@@ -0,0 +1,292 @@
...
+
+    # Drain remaining data from socket
+    while True:
+        ready, _, _ = select.select([s], [], [], 0)
+        if not ready:
+            break
+
+        buffer += s.recv(512)
+
+    return buffer
```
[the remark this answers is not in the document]
jmaksymowicz: Done

scripts/stm32n6-plo-rtt.cfg
```diff
@@ -0,0 +1,48 @@
...
+	# Wait a short time for plo and RTT to be initialized
+	sleep 500
+	halt
+	# On cold boot soon after resume the chip will reset and go back into Boot ROM.
+	# Check if we're executing out of Boot ROM now.
+	if { [lindex [get_reg {pc}] 1] < 0x20000000 } {
+		# After a reset something in the debug probe doesn't get reinitialized properly.
+		# For some inscrutable reason performing a byte or halfword (but not word!) access
+		# to any address makes the interface work again.
+		mdb 0x34180000
```
rmikielis, line 29: ugly hack with this mdb to reinitialize AP - is it the only way? how about ```reset halt```

jmaksymowicz: Unfortunately this doesn't work. If I replace this command with `reset halt` I get an error:
```
Error: timed out while waiting for target halted
TARGET: stm32n6x.cpu - Not halted
```

scripts/run-rtt-stm32n6.sh
```diff
@@ -0,0 +1,52 @@
...
+# Shell script for loading plo and launching RTT on STM32N6 devices.
+# It may be used for flashing or for running applications over RTT.
+#
+# Copyright 2026 Phoenix Systems
+# Author: Jacek Maksymowicz
+#
+
+set -e
+
+USAGE="Usage: $0 {flash,run} project_name"
```
rmikielis, line 12: <project_name>

scripts/run-rtt-stm32n6.sh
```diff
@@ -0,0 +1,52 @@
...
+else
+	echo "$USAGE"
+	exit 1
+fi
+
+openocd \
+	-f interface/stlink.cfg \
+	-f scripts/openocd/stm32n6/stm32n6x.cfg \
+	-c "adapter speed 24000" \
+	-f scripts/stm32n6-plo-rtt.cfg \
```
rmikielis, line 40: this command will require putting ```stm32n6-plo-rtt.cfg``` into openOCD location? Why not using project path?

scripts/run-rtt-stm32n6.sh
```diff
@@ -0,0 +1,52 @@
...
+USAGE="Usage: $0 {flash,run} project_name"
+
+PROJECT_NAME=$2
+PROG_PATH="_build/$PROJECT_NAME/prog.stripped"
+BOOT_PATH="_boot/$PROJECT_NAME"
+
+if [[ -z $PROJECT_NAME ]]; then
+	echo "$USAGE"
+	exit 1
+elif [[ ! $PROJECT_NAME =~ "armv8m55-stm32n6" ]]; then
```
rmikielis, line 21: if this script is for 1 target only, why <project_name> must be given as parameter? name of the script seems explanatory

jmaksymowicz: The script is intended for any project on target `armv8m55-stm32n6`. Currently we only have `nucleo` but you can make other projects for different boards or with different configuration (board config, Flash partitions, etc.)

Maybe the argument could be just the last component of the name (`nucleo` instead of `armv8m55-stm32n6-nucleo`). I was thinking about it, but we use the whole name everywhere else so I think this will be more consistent.

scripts/run-rtt-stm32n6.sh
jmaksymowicz: OpenOCD looks for a file in its `scripts` directory, then if it doesn't find it it looks in the current working directory - so we don't need to install anything into OpenOCD's location.

scripts/run-rtt-stm32n6.sh
jmaksymowicz: Done

jmaksymowicz: Done

